PR #96, opened 2014-12-24
Move browserify packages to dependencies

xdissent:
Currently the user requiring the lock package has to install them.


pose:
Hi @xdissent,

We've discussed your issue in the Auth0 team and we think that the best approach is to add these dev dependencies in your project:

```json
{
  "devDependencies": {
    "brfs": "0.0.8",
    "ejsify": "0.1.0",
    "packageify": "^0.2.0"
  }
}
```

We would like to prevent you from doing the extra work but we don't think that including these dependencies with the rest of the *runtime dependencies* is the way to go. Also, [we've updated our installation guide](https://github.com/auth0/lock/blob/master/README.md#browserify).

BTW, we are not a 100% happy with this solution and feel free to re-open if you find another way of solving this problem.

Thanks in advance,

PR #105, opened 2015-01-20
Add "Exact" logo to Lock

vctrfrnndz:
Adds exact logo and a README into how to update the icon font with new social providers. Also adds the latest icomoon build in /support folder.


lib/css/zocial.less
```diff
@@ -240,9 +242,8 @@ button:-moz-focus-inner {
...
-    url('https://cdn.auth0.com/w2/font/zocial-regular-webfont.ttf') format('truetype'),
-    url('https://cdn.auth0.com/w2/font/zocial-regular-webfont.svg#zocialregular') format('svg');
```
woloski: why removing this one? does it work without this on ie9?

vctrfrnndz: Mostly because they are outdated. All browsers except ie8 support woff. I also added the latest .ttf in base64 as a backup. Should be fine.

PR #114, opened 2015-01-30
Update fr-FR.json

bonatoc:
Updated to reflect the lastest en.json. 
When default language is switched to french, it causes an issue with ionic automatic generated sample downloadable from auth0.com, and probably with any configuration using Lock :

capture d ecran 2015-01-30 a 04 01 35

capture d ecran 2015-01-30 a 04 01 13

The error should be "silent", falling back to english if a string is not present in the targeted language. 
As it is, Lock just crashes.


jfromaniello:
Yes, it was falling back to english, it seems a regression. Thanks for reporting this

PR #116, opened 2015-02-03
Fixes #115: Wrong error message on sign up with sso domain.

### LINKED ISSUE #115: wrong error message on sign up with sso domain
It should show this message instead:
https://github.com/auth0/lock/blob/master/i18n/en.json#L45

pose:


jfromaniello:
Well done!

PR #118, opened 2015-02-04
Added recaptcha support

tehsis:
I've added recaptcha supports.
It expect an error property on login that will enable recaptcha v2.

Working example with mocks at http://codepen.io/tehsis/full/RNLbWJ

### DIFF
--- a/lib/mode-signup/index.js
+++ b/lib/mode-signup/index.js
@@ -125,13 +125,8 @@ SignupPanel.prototype.resolveOptions = function(options) {
 SignupPanel.prototype.bindAll = function() {
   var options = this.options;
 
-  // hide only and only if set to false
-  this.query('.a0-options')
-    .toggleClass('a0-hide', !options.showSignupAction);
-
   var list = this.query('.a0-iconlist').html('');
-  _.chain(options.$client.strategies)
-    .where({ social: true })
+  _.chain(this.options._getSocialStrategies())
     .map(function (s) {
       var e = {
         use_big_buttons: options._useBigSocialButtons(),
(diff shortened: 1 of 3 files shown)


lib/mode-signin/index.js
```diff
@@ -43,6 +44,8 @@ function SigninPanel(widget, options) {
   this.options = this.resolveOptions(options);
   this.el = null;
 
+  this.recaptcha = options.recaptcha?new Recaptcha(options.recaptcha):null;
```
pose: Can you add spaces between the operators? Thanks

test/tests.js
```diff
@@ -383,6 +383,32 @@ describe('Auth0Lock', function () {
       .show({ callbackURL: callbackURL, responseType: 'token', authParams: { state: 'foo' }});
     });
 
+   it.skip('should require a recaptcha token when login fails three or more times', function(done) {
```
pose: The test is skipped. Is this ok?

lib/recaptcha/index.js
```diff
@@ -0,0 +1,81 @@
...
+ * @param {NodeElement} el
+ * @param {Function} verifycb
+ * @return {RecaptchaWidget}
+ * @public
+ */
+
+Recaptcha.prototype.show = function(el, verifycb) {
+  this.verification = null;
+  if (null === this.widget) {
+    this.widget = grecaptcha.render(el, {
```
pose: What happens with `grecaptcha` object when there are multiple of Recaptcha in a page? Is there a way to prevent that object to leak as a global? Maybe there is no way, I'm just asking to understand.

lib/mode-signin/index.js
```diff
@@ -212,6 +215,33 @@ SigninPanel.prototype.bindAll = function() {
...
+    var recaptcha_token_holder = this.query('input[name=recaptcha-token]');
+
+    submit_button.attr('disabled', true);
+    
+    this.recaptcha.show(recaptcha_container[0], function(token) {
+      recaptcha_token_holder.val(token);
+      submit_button.attr('disabled', false);
+    });
+  } else {
+
```
pose: `else` can be removed or here should go recaptcha hiding logic?

lib/mode-signin/index.js
```diff
@@ -212,6 +215,33 @@ SigninPanel.prototype.bindAll = function() {
...
+  if (required) {
+    var self = this;
+
+    var recaptcha_container = this.query('.a0-recaptcha');
+    var submit_button =  this.query('button[type=submit]');
+    var recaptcha_token_holder = this.query('input[name=recaptcha-token]');
+
+    submit_button.attr('disabled', true);
+    
+    this.recaptcha.show(recaptcha_container[0], function(token) {
```
pose: What happens when `grecaptcha.render` fails? Is there a way to recover from that by making the function receive an `err` argument as well?

pose:
Sorry for so many comments. After addressing those items, looks good to me :+1: . Thank you very much for your hard work on this.

pose:
BTW, feel free to reach my by chat if I can help you with the changes or anything else.

lib/recaptcha/index.js
tehsis: multiple recaptchas is not a problem. Each instance generates its own id (asigned to this.widget here).

lib/mode-signin/index.js
tehsis: Yes, I thought on hiding logic. I'll fix the mistake

lib/mode-signin/index.js
tehsis: good question. I thought the same thing, but the docs didn't talk about that situation. I'll research a little about it. 
Anyway, I think it would be a really uncommon situation and a critic issue to Google. I mean, the recaptcha's script gets loaded but the render methods fail.

lib/recaptcha/index.js
pose: :+1:

PR #137, opened 2015-03-03
Added recaptcha support

tehsis:


ehough:
Is this PR still alive and considered for merging? I'm interested in this functionality and would be willing to contribute, but a quick status update first would be appreciated.

woloski:
Hi @ehough, thanks for the offer. This was a PoC. It requires a piece on the server side which we haven't done (check the captcha). So in short, you can work on the client side pieces to bring this to latest, but need the server side piece to check the captcha token.

gnandretta:
Closing because if we decide to add this will, we'll add it to v10 and that would require a rewrite.

lefty313:
@gnandretta 

because you are charging for successful login without captcha it's really easy to pay a lot for bots sign in. We are waiting for this feature since more than a year is there any `ETA` when it will be finally shipped ?

As a customer getting response like `well we rewrote our client so we don't care for this year old p/r` is not very helpful 👎

gnandretta:
Hi,

I closed the PR because it was a proof of concept that we didn't move forward. The reason is that we didn't have enough requests from our customers for this feature.

If you need captcha support our recommendation is to build a custom sign up with [auth0.js](github.com/auth0/auth0.js). But keep in mind that Auth0 doesn't charge you for successful login, it charges for regular active users.

Issue #188, opened 2015-06-27
Callback not called on widget close.


Zoltu:
```js
lock.showSignin(options, (error, profile, idToken) => {
	console.log("called");
}
```
If I click the x in the top right corner of the popup then "called" is not logged.

I'm using the GitHub release 7.5.4.

Zoltu:
https://github.com/auth0/lock/blob/master/index.js#L1272
https://github.com/auth0/lock/blob/master/index.js#L490
On those lines, `hide` is called with no parameter.

https://github.com/auth0/lock/blob/master/index.js#L490
On that line we can see that `hide` expects a `callback` parameter.

https://github.com/auth0/lock/blob/master/index.js#L505
Without the callback passed into `hide`, this function will never call the callback so the application won't be notified.

This is problematic for me because I can't tell the difference between the user waiting with the dialog open and the user closing the dialog.  If they close the dialog without logging in then I need to give them some kind of feedback letting them know that they will need to authenticate to complete whatever action they attempted.  As it is now, if the user doesn't log in then the page just stalls, assuming the lock popup is still open.

cristiandouce:
The callback parameter is not expected, it's just optional as you well noted below in here: https://github.com/auth0/lock/blob/master/index.js#L505.

Right below that we emit a `hidden` event which is emitted right after we invoke the optional callback.

If you want to get notified when the Lock is hidden, just do this:

```
lock.on('hidden', () => {
  console.log("lock hidden!")
})
```

The `show` callback is used to invoke on an authentication success/error. Not for the lock show/hide/ready events.

Zoltu:
From an API design standpoint, I believe it would be better if the callback passed to `show...()` was always called when Auth0 was done.

When I show the pop-up dialog, my application is effectively passing control to Auth0 and blocking while it waits for authentication to complete.  I need a reliable mechanism for knowing when Auth0 is done so I can proceed with my business logic (branching on whether the user is logged in or not).  Realizing that the 'hidden' event is the mechanism for knowing when I am done is very counterintuitive.  I expected to have my callback called by Auth0 when it was done doing "whatever it is that it does" but instead I only get that callback called in some situations.

At the least, this should be clearly documented since it goes against user expectations.

Handling this in my application is also more painful than if the callback were always called.

Code with callback getting called always:
```js
let authPromise = new Promise((resolve, reject) => {
	lock.showSignin(options, (error, profile, idToken) => (!error) ? resolve(profile) : reject(error));
}
```

[...]

PR #189, opened 2015-06-28
Fixes a bug where closing the popup dialog wouldn't call the callback.

Zoltu:
There may be a better way to do this, if so please let me know and I can update the PR.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1269,7 +1269,10 @@ Auth0Lock.prototype.getProfile = function (token, callback) {
 
 Auth0Lock.prototype.oncloseclick = function(e) {
   stop(e);
-  this.hide(() => this.options.popupCallback(new Error("Dialog closed.")));
+  self = this;
+  this.hide(function() {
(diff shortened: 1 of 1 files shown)


Zoltu:
Fixes #188

index.js
```diff
@@ -1280,7 +1283,11 @@ Auth0Lock.prototype.oncloseclick = function(e) {
  */
 
 Auth0Lock.prototype.onescpressed = function(e) {
-  if ((e.which == 27 || e.keycode == 27)) this.hide();
+  self = this;
```
cristiandouce: A few comments:

1. `self = this;` missing `var` on this line.
2. There may be times where the popupCallback is undefined, which means that invoking it will crash the script.
3. I'm not sure an error is appropiate here, since it's not an authentication error. We do have an event called `hidden` which is invoked when the Lock is hidden.

cristiandouce:
Actually, yeah... I think this is not required. Thanks though for the initiative!

you can use the events to be notified when the lock is hidden.

Issue #183, opened 2015-06-23
SSO UX is wrong in the Sign Up and Reset Password views


sandrinodimattia:
For Enterprise Connections like ADFS, Azure AD, ... the Lock supports HRD based on the Email Address of the user.

As a result, the Sign In view will show the SSO login button when the email address matches one configured on one of the connections:

![image](https://cloud.githubusercontent.com/assets/655409/8303598/426db7d0-199f-11e5-8ca2-b27eda8bd970.png)

While this is working great for sign in, users are still confused in some situations. This is why we should take the following into account when we have a user enabled for SSO:

 1. Hide the Sign Up and Reset Password buttons when we show the SSO login button. Show them again as soon as the email address is no longer SSO enabled.
 2. When users enter the Sign Up or Reset Password views and enter their email address they think they can for example change their password from here, which is not the case. When they enter an SSO enabled email address, the Lock should go back to the Sign In view with that same Email Address already filled in and the SSO login button visible.

lilliann123:
Centralpoint really is an extraordinarily helpful CMS compared to the other ones out there.  The single sign on piece lets me integrate right into ldap/ad to organize the sites users from the get go. I can export records from the system as word docs or pdfs right on the fly. The document management system lets me create 'portions' in the document that are dynamic, so when one person signs in they get the document they need vs. someone else.

hzalaz:
Already fixed by #191

PR #191, opened 2015-07-01
Improve SSO experience

sandrinodimattia:
This will improve the SSO experience when a user enters their email address for which the email suffix has been configured on a connection:

 - Signup and Reset buttons are hidden when we are in HRD mode
 - If you enter an SSO enabled email address on the SignUp or Reset panel, the Lock will navigate back to SignIn, enter your email address there and trigger HRD
 - If you then edit your email address and no longer have an SSO enabled email address, you'll be redirected back to the previous panel (SignUp or Reset). This is required when we start the Lock in SignUp mode only for example

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,7 @@ require('./lib/insert-css');
  */
 
 var bonzo = require('bonzo');
+var bean = require('bean');
 var _ = require('underscore');
 var debug = require('debug')('auth0-lock');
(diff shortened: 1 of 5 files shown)


sandrinodimattia:
Fixes #183

cristiandouce:
:heart_eyes_cat: Will review later today! :guitar: :star:

PR #192, opened 2015-07-03
Added AD Usernaame Provider

sandrinodimattia:
Allow injecting your own logic to resolve the AD username. Eg:

```
            lock.show({
              adUsernameProvider: function(email) {
                var name = email.substring(0, email.lastIndexOf("@"));
                var domain = email.substring(email.lastIndexOf("@") + 1).replace('.com', '');
                return domain + '\\' + name;
              },
              callbackURL: window.location.href,
              responseType: 'token'
            });
```

As a result, instead of just showing the username in the AD login form, it will show: `contoso\jon`


gnandretta:
Closing because if we add this to v10 would require a rewrite.

Issue #82, opened 2014-11-14
Add more events for analytics purpose


cristiandouce:
> This would be a core step in the sign-up analytics funnel (initalise, ready, submit, complete)

**source:** https://ask.auth0.com/t/analytics-event-on-lock/299

cristiandouce:
@mikerkeating I'd love your feedback on this once it's released the latest version!

Thanks!

PR #196, opened 2015-07-20
Added Lifecycle events

tehsis:
### Briefing

This PR completes the lifecycle of events for our modes: `signin`, `signup`, `reset`.

- Close auth0/lock#82

The states added are: `ready`, `submit`, `success` and `error`.

Which will generate events like

```js
lock.on('signup ready', function () {})
lock.on('signup submit', function () {})
lock.on('signup success', function () {})
lock.on('signup error', function () {})

// ... same for signin and reset
```

### DIFF
--- a/lib/mode-loggedin/index.js
+++ b/lib/mode-loggedin/index.js
@@ -121,6 +121,7 @@ LoggedinPanel.prototype.bindAll = function() {
 
   this.query('form').a0_on('submit', function (e) {
     stop(e);
+    widget.emit('loggedin submit', widget.options);
     widget._signInEnterprise(e);
   });
 
@@ -146,6 +147,7 @@ LoggedinPanel.prototype.bindAll = function() {
   this.query('.a0-strategy .a0-zocial[data-strategy]').a0_on('click', function (e) {
     stop(e);
 
+    widget.emit('loggedin submit', widget.options, widget.$ssoData);
     widget._signinSocial(
       strategy_name,
       widget.$ssoData.lastUsedConnection && widget.$ssoData.lastUsedConnection.name,
--- a/lib/mode-reset/index.js
+++ b/lib/mode-reset/index.js
@@ -12,6 +12,7 @@ var bind = require('../bind');
 var template = require('./reset.ejs');
 var regex = require('../regex');
 var PasswordStrength = require('../password-strength');
+var ValidationError = require('../errors/ValidationError');
 var empty = regex.empty;
 var trim = require('trim');
 var email_parser = regex.email_parser;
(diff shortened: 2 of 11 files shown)


lib/mode-loggedin/index.js
```diff
@@ -146,6 +147,7 @@ LoggedinPanel.prototype.bindAll = function() {
   this.query('.a0-strategy .a0-zocial[data-strategy]').a0_on('click', function (e) {
     stop(e);
 
+    widget.emit('loggedin submit', widget.options);
```
cristiandouce: pass along lastUsedConnection and getLoggedInAuthParams... something like when triggered _signin

lib/mode-reset/index.js
```diff
@@ -221,34 +222,47 @@ ResetPanel.prototype.valid = function () {
   widget._focusError();
 
   if (email_empty) {
+    if (validate_username) {
+      widget.emit('reset error', new ValidationError('username empty'));
+    } else {
+      widget.emit('reset error', new ValidationError('email empty'));
+    }
```
cristiandouce:
```js
widget.emit('reset error', new ValidationError(validate_username ? 'username empty' : 'email empty'));
```
make more sense to me and avoids the `block hell`

lib/mode-reset/index.js
```diff
@@ -221,34 +222,47 @@ ResetPanel.prototype.valid = function () {
...
-      ok = false || (validate_username && username_parsed);
-      if(!ok) widget._focusError(email_input, widget.options.i18n.t('invalid'));
+    ok = false || (validate_username && username_parsed);
+
+    if (validate_username && !username_parsed) {
+      widget.emit('reset error', new ValidationError('username invalid'));
+      widget._focusError(email_input, widget.options.i18n.t('invalid'));
+    } else if (!ok) {
+      widget.emit('reset error', new ValidationError('email invalid'));
+      widget._focusError(email_input, widget.options.i18n.t('invalid'));
```
cristiandouce: same here... this can be improved, don't you think?

lib/mode-reset/index.js
```diff
@@ -221,34 +222,47 @@ ResetPanel.prototype.valid = function () {
...
 
   if (repeat_password_input.val() !== password_input.val()) {
+    widget.emit('reset error', new ValidationError('password missmatch'));
     widget._focusError(repeat_password_input, widget.options.i18n.t('mustMatch'));
     ok = false;
   }
 
   return ok;
-}
+};
```
cristiandouce: maybe we could save the `error = new ValidationError('...');` and trigger the emit at the end right before the `return ok`.

lib/mode-signin/index.js
```diff
@@ -349,11 +351,19 @@ SigninPanel.prototype.onsubmit = function(e) {
...
+      } else if (!ok) {
+        widget.emit('signin error', new ValidationError('email invalid'));
+      }
+
+      if (!ok) { widget._focusError(email_input, options.i18n.t('invalid')); }
     }
   }
 
   if (password_empty && password_required && !password_disabled) {
+    widget.emit('signin error', new ValidationError('password empty'));
```
cristiandouce: same as reset... save the Error and emit on the end of this validate method

lib/mode-signup/index.js
```diff
@@ -374,16 +386,19 @@ SignupPanel.prototype.valid = function() {
...
     }
     if (!username_parsed && !username_empty) {
+      widget.emit('signup error', new ValidationError('username invalid'));
       widget._focusError(username_input, widget.options.i18n.t('invalid'));
       ok = false;
     }
   }
 
   if (password_empty) {
+    widget.emit('signup error', new ValidationError('password empty'));
```
cristiandouce: same as signin and reset, save the error and emit at the end since we are not returning fast in either `.validate` method.

tehsis:
Done :+1: 

We've decided not to move the events so we can fire up different events (e.g. email invalid and password empty) and be consistent with focuserror.

lib/mode-reset/index.js
```diff
@@ -294,12 +305,14 @@ ResetPanel.prototype.submit = function () {
...
       widget._signinPanel(panel.options);
       widget._showSuccess(widget.options.i18n.t('reset:successText'));
-
+      widget.emit('reset success');
       return 'function' === typeof callback ? callback.apply(widget, args) : null;
     }
 
     widget.setPanel(panel);
 
+    widget.emit('reset error', err);
```
cristiandouce: here we are triggering the error before the setPanel which will trigger the error after the ready...

different from what we established, right?

lib/mode-signin/index.js
```diff
@@ -382,6 +390,7 @@ SigninPanel.prototype.onsubmit = function(e) {
 
 SigninPanel.prototype.onsocialclick = function(e) {
     stop(e);
+    this.widget.emit('signin submit', this.widget.options);
```
cristiandouce: signing submit with what exactly? google? twitter?

lib/mode-reset/index.js
tehsis: Right, will change.

tehsis:
I've checked the events and are ok from my pov. I think we can improve Kerberos, but I'll recommend to do it in a separate issue since there seems to be  a lot of unused code and some things that needs to be refactored.

PR #213, opened 2015-09-03
popup property overriden by sso:true && popupCallback:{Function}

cristiandouce:


lib/mode-signup/index.js
```diff
@@ -270,7 +270,7 @@ SignupPanel.prototype.submit = function() {
   // IMPROVE: This is a hack to allow `loginAfterSignup` with `sso: true`
   // since the auth0-js client requires a popup for that, and it needs to
   // be open in the same thread as the event emitted by the user action
-  if (panel.options.loginAfterSignup) {
+  if (panel.options.loginAfterSignup && options.popup) {
```
glena: mm `options.popup` or `options.sso`

cristiandouce: I can have `options.sso = true` but be in a redirect mode. For that case i dont want the popup open. That's the reason for the double check in options, I force `popup = true` only if sso and callback (no-redirect mode)

PR #217, opened 2015-09-17
pl i18

rpdiss:
added Polish locale


cristiandouce:
Thanks! will release soon

cristiandouce:
@rpdiss already released as `lock@7.9.4`. Thanks again!

PR #229, opened 2015-10-02
Improve webpack instructions

taion:
I actually didn't find the webpack instructions because I assumed there weren't any after I scrolled past the Browserify section, and ended up wasting a little bit of time.

This improves the organization of that section of the README a bit and will hopefully prevent other people from making the same mistake in the future.

This also improves the webpack example a little bit by restricting the scope of the transforms to just the `auth0-lock` package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ If you are targeting mobile audiences, it's recommended that you add:
 
 ### Browserify
 
-If you are using browserify to build your project, you will need to add the following transformations required by Auth0 Lock:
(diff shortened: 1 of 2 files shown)


cristiandouce:
Awesome, but could you please restore the examples section?

taion:
Done.

My original reasoning was just that it seemed more like just webpack-related documentation than a standalone example per se.

cristiandouce:
Thanks!

PR #235, opened 2015-10-07
Use cacheable loaders in webpack

taion:
This speeds up incremental builds in webpack for users a bit; previously the files were not cacheable, and had to be re-built by webpack on incremental builds even if they didn't change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,9 +77,9 @@ If you are using webpack, you will need the following loaders:
 [
   {
     test: /node_modules\/auth0-lock\/.*\.js$/,
-    loaders: 'transform?brfs!transform?packageify'
+    loaders: 'transform/cacheable?brfs!transform/cacheable?packageify'
   },
-  {test: /\.ejs$/, loader: 'transform?ejsify'},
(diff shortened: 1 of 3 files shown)


cristiandouce:
Though I'm moved by perf and faster build processes I'm also concerned of the issues it might bring using 2 different libs at the moment of building the same templates.

I see you are changing the `ejsify` lib for `ejs-compiled-loader`.

taion:
They both just call into https://www.npmjs.com/package/ejs anyway. Although `ejsify` uses 1.x and `ejs-compiled-loader` uses 2.x. That doesn't make a practical difference here.

cristiandouce:
We'll see :dancers: 
If troubles appear, I guess we could push `ejsify` to move to `ejs@2.x`, right?

taion:
Or downgrade `ejs-compiled-loader` to `1.x`.

The differences are pretty minor. `2.x` takes out support for filters, which you don't use. It adds a few features like per-template custom delimiters and run-time includes. I don't think the former can be used with the current setup, and the latter really work when bundling anyway, and I don't think either are likely to be especially useful here anyway.

Issue #225, opened 2015-09-29
Unable to build with browserify (nodejs 4.1, npm@3)


adrianhall:
Utilizing gulp + browserify, I get an error when building my app that includes auth0-lock

nodejs 4.1.1, npm 3.3.3, browserify 11.2.0, babelify 6.3.0, auth0-lock 7.9.5, brfs 1.4.1, ejsify 1.0.0, packageify 0.2.2

Build process is in gulp: 

```
function () {
    var b = browserify({
        entries: config.src.jsBundle,
        extensions: ['.js', '.jsx'],
        transform: [ babelify.configure(config.options.babel) ],
        debug: true
    });

    return b.bundle()
        .pipe(source('bundle.js'))
        .pipe(buffer())
        .pipe(sourcemaps.init({ loadMaps: true }))
        //.pipe(uglify(config.options.uglify))
        //.on('error', gutil.log)
        .pipe(sourcemaps.write('.'))
        .pipe(gulp.dest(config.dest.path));
}
```

The code compiles cleanly when I don't `require('auth0-lock');`.  As soon as I include auth0-lock, I get a compile error:

``` 
events.js:141
      throw er; // Unhandled 'error' event
      ^

SyntaxError: Unexpected token (1:0) while parsing file: C:\Users\ADRIANHA\Documents\GitHub\adrianhall\GrumpyWizards\grumpy-wizards\src\Website\node_modules\auth0-lock\lib\html\main.ejs
    at Parser.pp.raise (C:\Users\ADRIANHA\Documents\GitHub\adrianhall\GrumpyWizards\grumpy-wizards\src\Website\node_modules\acorn\dist\acorn.js:1745:13)
    at Parser.pp.unexpected (C:\Users\ADRIANHA\Documents\GitHub\adrianhall\GrumpyWizards\grumpy-wizards\src\Website\node_modules\acorn\dist\acorn.js:2264:8)
    at Parser.pp.parseExprAtom (C:\Users\ADRIANHA\Documents\GitHub\adrianhall\GrumpyWizards\grumpy-wizards\src\Website\node_modules\acorn\dist\acorn.js:1141:12)
    at Parser.pp.parseExprSubscripts (C:\Users\ADRIANHA\Documents\GitHub\adrianhall\GrumpyWizards\grumpy-wizards\src\Website\node_modules\acorn\dist\acorn.js:1023:19)
[...]

Issue #7, opened 2014-08-23
Improve UI customization wiki documentation


cristiandouce:
https://github.com/auth0/lock/wiki/UI-Customization

@gravityonmars anything could be better than that.

Also a few screenshots might be awesome! :dancer:

PR #236, opened 2015-10-07
Proposal: Distribute bundle on npm as well

taion:
Don't merge this as is - I just wanted to drop this PR as an example of the non-documentation changes involved.

It doesn't seem like there's a meaningful reason to distribute this on NPM as source rather than as a built bundle.

- There don't seem to be any hooks other than the main import, so I'm not doing any partial imports anyway
- It significantly complicates building by requiring users to add extra transforms for Browserify and extra loaders for webpack
- Potential additional minification benefits are not material; besides, at > 0.5 MB, anybody who really cares is going to be using code splitting on this package anyway

If you distribute this as the bundle, then Browserify and webpack users can drop all the extra build steps.

Full disclosure, on webpack doing this will result in the following unavoidable warning:
```
WARNING in ./~/auth0-lock/build/auth0-lock.js
Critical dependencies:
1:113-120 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
 @ ./~/auth0-lock/build/auth0-lock.js 1:113-120
```

This is less of an issue to me than adding a bunch of extra loaders that don't tie into my app at all.

Also, for actual distribution you'd probably want a NPM-specific bundle that doesn't bundle in e.g. auth0.js and your other dependencies.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -15,8 +15,5 @@ results
 npm-debug.log
 node_modules
 
-#do not include bundle
-build
-
 # other paths to ignore
 bin
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "auth0-lock",
   "version": "7.9.5",
   "description": "Auth0 Lock",
+  "browser": "build/auth0-lock.js",
   "main": "./index.js",
   "keywords": [
     "auth0",


cristiandouce:
Interesting, I'll take a look.

We already had this discussion over [auth0.js](https://github.com/auth0/auth0.js) in the past and we dropped the idea of having the build distributed over npm registry.

We also offered an alternative for using npm + having the build through github archives:

Take a look at it here: https://github.com/auth0/auth0.js/issues/91#issuecomment-107631673

taion:
Looks like you guys dropped `packageify` there, though.

As always there are tradeoffs, but it feels strange to have to set up so many loaders or transforms to get things to work. Ideally I want my build to, as much as possible, only cover my own code.

For example, projects that use ES6 will transpile their code with Babel and publish only the transpiled code to NPM, instead of having all users use Babel as well.

If it were easily possible, the right approach might be to pre-transform the individual files (i.e. to still use CJS, but to pre-resolve non-JS imports), but I don't know how to do that.

cristiandouce:
@taion I think you missed the point there.

If you want `npm` + `build/auth0-lock.js` you can `npm install auth0/lock#7.9.0` and that package will have the build bundle to use.

taion:
Understood. What I'm saying is that other Browserify/webpack users going forward would have an easier time at it if `package.json` pointed them at something that already had EJS, packageify, and the CSS `readFileSync` already processed.

The current install workflow looks something like -

1. Install `auth0-lock` from NPM
2. Read the first page or so of docs, then try to use it
3. My build breaks
4. Go back and find the Browserify or webpack instructions in the docs and follow those

It might be nice if others could skip steps (3) and (4), because they are a little bit annoying (but not showstoppers). This is one way to do so, though it's probably not ideal.

LegNeato:
Totally agree. NPM packages generally target the lowest common denominator.

chrisco255:
This has been difficult to integrate with the Webpack workflow even with the given instructions.

jfromaniello:
Why the build breaks when you install from npm? Is it because of the transformer?

theopak:
@chrisco255 yeah, it was mildly annoying that the webpack example didn't work out of the box. Mind sharing your fixes? I was able to add auth0-lock to my project with just a couple of changes (see #251).

theopak:
My package.json has this dependency: `"auth0-lock": "auth0/lock#v7.10.3"`

taion:
How does the webpack example not work out of the box? If you go into that directory and run, it worked, last I checked.

theopak:
@taion I copy-pasted the code from the readme into my project, and it didn't work. The fix is very easy: #251

baer:
I'd like to +1 this for three reasons:

1. In corporate projects, asking to modify the build process is a non starter. One of the projects I work on has 150k+ lines of javascript across a dozen or more teams. The build is very sensitive and it would be a non starter to ask to modify it for a vendor dependency.

2. I in both personal projects and at work use package.json is to keep track of dependencies. If I am forced to install at a SHA I am much less likely to keep up to date since npm doesn't do a very good job of dealing with version bumps when you depend directly on a github/bitbucket project.

3. It creates a lot of friction for developers who want to use your awesome project!

taion:
@baer 

To play devil's advocate a bit, though, it's not uncommon for libraries that ship styled components to ask users to do some amount of work to pull in e.g. the CSS or whatever.

To the extent that users want to customize the CSS used here, bundling it into the JS is not necessarily an ideal option.

I still think on balance this package should deliver a directly-import-able bundle, but maybe the real problem is that webpack doesn't have a concept of "this module needs these transforms to work", the way that e.g. browserify does (obviously not a relevant point if you're not a webpack user).

baer:
This lib has provided two (blessed) ways to integrate the library and what I think I'm asking for is something in the middle. On the one side you have the CDN distributed file that is fully compiled, minified and ready to go. On the other side you are exposing your source code to such an extent as to require a developer to modify her/his source code.

I generally agree with exposing your source like this BUT requiring people to modify their builds is a hard pill to swallow. One middle ground option that would satisfy folks like me would be to provide a bundled (but NOT minified) CSS/JS version that we could pull into our build process and manipulate as needed. One caveat here is that Webpack SUCKS at consuming other Webpack builds because of how it's module registry works.

jfromaniello:
@baer  another issue with this change is that you will install all the dependencies even if you don't need them.. Maybe we can modify our release pipeline to push something to npm as `lock-bundle` or `lock-webpack`.

With the changes shown in this PR, plus removing all dependencies. Will that work for you guys?

baer:
I just discovered another issue with not providing a dist bundle: https://github.com/auth0/lock/issues/225

It looks like your build also has to be running npm@2.x

ianks:
Building `auth0-lock` with webpack is very slow. I would love to have a bundle on NPM that I could use instead of having to deal with compiling ejs, etc.

ashishtilara:
+1

yangmillstheory:
+1

101100:
Many packages offer the minified version of their code inside the npm bundle.  This allows the use of both direct imports and also a bundle for those that want that but don't want to use the CDN.  Often multiple options are available (minified, with and without source maps, etc).

taion:
It looks like this PR isn't going to get merged, so I'll close it out to reduce noise.

Issue #244, opened 2015-10-21
Lock re-login process flow stuck


lordakshaya:
After I have logged into my application once, when I open the app again - I get stuck on this page:

![image](https://cloud.githubusercontent.com/assets/1169337/10645470/119e9e04-7858-11e5-89b0-e8791a0e323d.png)

rolodato:
Any JS errors or failed requests in dev tools? A sample project to reproduce the problem would also help.

lordakshaya:
@rolodato:
Console Error - none

Response JSON structure: 

```
lastUsedClientID: "xxxxxxxxxxx"
lastUsedConnection: {name: "sms", strategy: "sms"}
lastUsedUserID: "sms|xxxxxxxxxx"
lastUsedUsername: "+0000000000"
sessionClients: ["xxxxxx"]
sso: true
```

Clicking on Show all, doesnt trigger anything.

gnandretta:
@lordakshaya @rolodato We've released v7.10.2 which should fix the issue.

rolodato:
@lordakshaya Just to give some context, this was being caused because Lock fetches the Auth0 SSO cookie to determine which connection was used by the user. Lock was not recognising `sms` as a valid connection, so it failed to display the dialog. 7.10.2 was updated to ignore unknown connections instead of failing.

PR #246, opened 2015-10-21
Don't show last login for unknown connections

gnandretta:
When the connection is unknown and SSO is enabled `LoggednPanel` is displayed without any buttons, as reported in #244.


gnandretta:
@cristiandouce could you review the change please?

cristiandouce:
Perfect!! :smile: 
merge and patch release :dancers:

PR #253, opened 2015-11-10
Show which loaders to install in README

ianks:


gnandretta:
Thanks! I've already merged #251 which documents needed loaders in the README and improves the webpack example.

ianks:
@gnandretta That example is still wrong. It needs to be `loader` singular for the first one.

ianks:
you could cherry pick https://github.com/ianks/lock/commit/6296c9616cbca6116f1e0444f7eb414041402d53 for that.

gnandretta:
@ianks The commit ianks@86a75dc is outdated. It should be fixed in the README now.

Issue #230, opened 2015-10-02
Elements not centered on Chrome 47


taion:
Running `47.0.2522.0 (Official Build) dev (64-bit)` (dev channel). On the playground (and elsewhere) I see:

![image](https://cloud.githubusercontent.com/assets/3112159/10238700/905a325c-6891-11e5-83f7-3f388e0d99c0.png)

The `text-align: center` styling is not doing anything.

This doesn't happen on Chrome 45. I don't know for sure that this isn't a Chrome bug, but I haven't noticed this sort of thing elsewhere.

taion:
Still broken on `48.0.2541.0 (Official Build) dev (64-bit)`.

taion:
This is also broken on `47.0.2526.16 (Official Build) beta (64-bit)` (current beta channel) and on the current Canary channel release of Chrome as well.

glena:
Hi @taion, we only provide support up to the latest stable version.

Is hard to provide compatibility with beta/canary versions since they iterate too fast and is likely to have js and rendering issues from time to time.

Anyway, there shouldn't be any issue with stables releases.

taion:
@glena The problem is - if this is actually a bug, and the rendering behavior on Chrome beta is representative, then every website using the current release of `auth0-lock` will show a broken login modal for their users as soon as Chrome 47 hits stable. The fix for your users here wouldn't just be to update this package, it'd be for you to update this package and for your users to update their dependent versions. I really don't want to start showing a broken login modal to Chrome users when Chrome 47 hits stable.

glena:
Non stable releases have this kind of problems from time to time and usually it is related to some internal issue on the rendering engine.

Since it is a very simple layout, it's a standard css, there is no crazy stuff going on that could make this kind of problems.

cristiandouce:
Besides, if this comes to be a real issue in the future, users still will have to update the `auth0-lock` version to the one containing this fix. Either they do it now (assuming the bug will be released out in Chrome), or later when Chrome 47 is out.

What doesn't make sense IMHO is to force now the update for a bug that might or might not exist.

gavinharriss:
This same issue has now manifested again with the newly released 47.0.2526.73 version of Chrome.

glena:
and it is fixed with 7.12.3

gavinharriss:
Thanks, I didn't realise I was requesting a specific version of the lock widget. Confirmed that everything works with the latest version.

PR #259, opened 2015-11-20
Fixes alignment in Chrome

gnandretta:
The issue was reported in #230.

By setting text-align-last to auto it will use the value of text-align (unless is justifty).

More info on https://developer.mozilla.org/en-US/docs/Web/CSS/text-align-last.


taion:
Thanks! I was pretty happy to see the alignment working correctly again.

Issue #257, opened 2015-11-17
Bug: Form doesn't reset to login after failed autoLogin after signup


baer:
If `loginAfterSignup` is `true` (the default behavior) and a user's login fails after signup the form values don't reset and the form stays in "signup" mode. In order to clear it you have to click the "cancel" button.

gnandretta:
I can confirm this is happening.

After a successful signup, [Auth0Lock._signinWithAuth0](https://github.com/auth0/lock/blob/d7dbd1d1738dd1f5ffa8ff18476fe445ba4fd767/lib/mode-signup/index.js#L308) is called with an instance of `SignupPanel` from which it queries the user name and password to log the user in. That same instance is also rendered when the login fails.

We'll need to schedule some time to tackle this because I couldn't find a quick solution.

baer:
Thanks for the quick response! I'm looking forward to having a working Auth0 pipeline in my app.

baer:
@gnandretta - Any updates on this? I can't roll out Auth0 until this is fixed :/

Workarounds:
* I could turn off autoLogin but your dashboard doesn't index users until they have attempted to sign in so this will result in our admin trying to manage a bunch of users that appear to not exist.
* I could show and hide Lock to reset it but then I lose all of the nice validation messages you provide which would be pretty confusing to a user.

Any ideas?

gnandretta:
@baer when do you plan to launch? I'll fix it tomorrow

baer:
I will launch when I get all of my auth stuff straightened out and it is now looking like I have a bit of backend work to do in order to expire a token. Whenever you get to it would be great but I won't launch until late this week or early next week it looks like.

gnandretta:
@baer should be fixed in _7.12.2_.

PR #261, opened 2015-11-24
Improve error handling when logging in automatically

gnandretta:
If something goes wrong during the login, the _sing in_ panel should be displayed instead of the _sign up panel_. See #257.


cristiandouce:
Looks good to me :+1:

PR #265, opened 2015-12-08
Update de.json - signup 

johannesboyne:
The english "Signup" should be translated "Registrieren" in German.


cristiandouce:
Thanks!

PR #267, opened 2015-12-17
Consider stage subdomain when building assets url

gnandretta:
A new `stage` environment has been added and Lock is failing to obtain the client settings from the CDN because the assetsUrl is incorrect.


cristiandouce:
:+1:

PR #279, opened 2015-12-28
Electron support

gnandretta:
Adds support for electron:
- Allows to override `Auth0.prototype.openWindow` (see the details in the PR linked above)
- Loads the default icon correctly when `window.location.protocol` is `"file:"`, which is the case for Electron.


cristiandouce:
:+1:

Issue #284, opened 2016-01-10
Repo License needs an update


aguerere:
Also the footer of the Readme.md is missing author and license info.

PR #285, opened 2016-01-10
Update license and readme

aguerere:
Adds author and license info to the readme. Updates license file.

Fixes https://github.com/auth0/lock/issues/284

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,6 @@
-
 The MIT License (MIT)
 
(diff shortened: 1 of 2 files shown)


gnandretta:
Thank you!

aguerere:
My pleasure

jfromaniello:
I think it should be 2014 and not 2016 as it was before, why did you change that?

PR #286, opened 2016-01-11
Fixes

gnandretta:
- Fix flashing lock when in popup mode
- Remove .mode-container background color: the border radius of this div wasn't playing nice with the border radius of the submit buttton and a the edge looked jagged.
- Update font family


cristiandouce:
:+1:

PR #289, opened 2016-01-13
Threat browserify transforms as regular dependencies

gnandretta:
This will make it easier to bundle the npm module with browserify since it doesn't require any extra step.


cristiandouce:
:+1:

jfromaniello:
YES! :+1: this has been broken since its conception

taion:
This is breaking all of your webpack users. It's the root cause of https://github.com/auth0/lock/issues/301 and https://github.com/auth0/lock/issues/302, because you're pinning absolutely ancient versions of the relevant Browserify transforms.

This is also conceptually wrong – these are exactly development dependencies and should be treated as such.

PR #295, opened 2016-01-21
Fixes

gnandretta:
- Avoid flicker when removing avatar (Fixes auth0/lock/issues/292)
- Fix firing of avatar and icon events
- Show error if no valid connection could be found (Fixes auth0/lock/issues/14)


cristiandouce:
:+1:

PR #300, opened 2016-01-25
Improved webpack example

maoueh:
* Make it work on Windows by changing matching rules to use a broader
   regex but narrower include path(s).
* Scope JSON loader to auth0 packages only so impact is minimal when
   there is other JSON files in a project which should not be processed.
* Bumped auth0-lock dependency to 8.x line to be more up-to-date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,17 +65,20 @@ If you are using Webpack, you will need to install loaders (`$ npm install --sav
 
 ```js
 loaders: [{
-  test: /node_modules\/auth0-lock\/.*\.js$/,
+  test: /\.js$/,
   loaders: [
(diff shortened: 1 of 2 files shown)


maoueh:
@gnandretta I rebased my PR on master but I had modified slightly differently because I thought that transforming `ejs` file was not required anymore. I re-tested the rebased version with what the README.md was actually saying and it seems to work.

I will do some more validation just to be sure.

llops:
Wow, thank you very much @maoueh!! I lost 2 hours trying to run the original example on Windows... 
I hope Auth0 merge your PR soon.

gnandretta:
We have released v10 which doesn't require any special loader.

Issue #301, opened 2016-01-25
8.2.0 webpack not working due to unresolve 'fs' module


maoueh:
The version `8.2.0` does not work with webpack anymore. The example will fail with error message:

```
ERROR in .\~\auth0-lock\lib\insert-css\index.js
Module not found: Error: Cannot resolve module 'fs' in \work\auth0\lock\examples\webpack\node_modules\auth0-lock\lib\insert-css
 @ .\~\auth0-lock\lib\insert-css\index.js 5:9-22
```

To replicate, simply edit `examples/webpack/package.json` and change `"auth0-lock": "^7.5.5",` to `"auth0-lock": "^8.0.0",` and then issue the command to test the example (`npm install && npm start`). You will get the error mentionned above.

If you fix the version to `8.1.5` (via `examples/webpack/package.json` line `"auth0-lock": "8.1.5",`) error is not present.

The module in question (`lib\insert-css\index.js`) has a `require('fs')` line since a while, way before `8.1.5` which is working. This would mean that before, this module was not directly referenced while it is now in `8.2.0`. 

I will need further investigation to find the actual culprit but I need to step out for now.

Regards,
Matt

gnandretta:
@maoueh sorry for the inconvenience, could you try with version `8.2.1`? it should be fixed

maoueh:
@gnandretta Yep, just tested `8.2.1`, works like a charm.

PR #304, opened 2016-01-26
Fix webpack bundle

gnandretta:
- Upgraded brfs version, because it doesn't work with webpack (see #301). Since lock 8.2.0 it's included as a regular dependency and is the one that webpack picks up.
- Upgraded lock version in webpack example and updated instructions in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ No special configuration is required if you are using Browserify to build your p
 
 ### Webpack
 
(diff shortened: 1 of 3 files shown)


daf-spr:
LGTM :+1:. Tested locally it works ok, just fix what we talked about and LGTM

gnandretta:
thanks @daf-spr !

maoueh:
@gnandretta The readme seems to still reference old loaders.

PR #310, opened 2016-02-04
Don't optimize legibility for social icons

gnandretta:
It seems this sometimes makes Chrome render the icon without the zocial font when the lock is shown as soon as the page loads.

screen shot 2016-02-04 at 2 58 13 pm


beneliflo:
Looks good to me :+1:

PR #315, opened 2016-02-12
XHTML-compilance.

rafalwrzeszcz:
**Lock** library is right now totally inusable in XHTML applications. XHTML enforces some rules that don't take place in HTML mode:

- each element must be closed (at least self-closed);
- each attribute must have value;
- all text must be properly escaped.

I included XHTML test runner, so tests against HTML and XHTML can be run.


gnandretta:
Hi, we are focusing on adding new features only to v10 and I'm not sure if doing this will be possible because we are using React to manipulate the DOM.

PR #321, opened 2016-02-25
Fixes

gnandretta:
- Replace reset password flow (although discouraged, legacy flow is still available by passing the `useNewReset: false` option)
- Use a more descriptive message when too many requests are made to reset the password (for both flows)
- Don't run callback passed to `showReset` when resetting the password. Callbacks are only used for auth, for other things use an event.
- Fix issue in social buttons for idps with a long name when using the "es" dict.
- Fix success message styling.
- Add missing translations to some dicts.


cristiandouce:
LGTM :+1:

PR #331, opened 2016-03-08
Custom avatars

chazu:
This PR adds a runtime option to supply custom logic for avatar URLs - useful for those users whose platforms include their own avatars.

In order to utilize the functionality, the lock widget can be instantiated using a 'fetchAvatar' function which receives the widget instance and the email field value. This function is expected to operate analogously to the gravatar function supplied with lock - that is, it is expected to call widget.setImage() with an appropriate URL.

Please let me know if you have any questions, corrections or suggestions which would help make this patch suitable for inclusion in mainline lock.

### DIFF
--- a/lib/avatar/index.js
+++ b/lib/avatar/index.js
@@ -1,0 +1,30 @@
+var md5 = require('../md5');
+var regex = require('../regex');
+var trim = require('trim');
+
+/**
+ * Returns a function which sets the header image using the supplied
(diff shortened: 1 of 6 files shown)


gnandretta:
@chazu this look awesome! I'll review it during the week

pklingem:
excited for this!

chazu:
Closing this since @gnandretta reimplemented this for lock 10.

PR #377, opened 2016-05-08
add bitbucket and dropbox social connections - starting point

arcseldon:
This is a starting point only - the CSS and SVG logo selection etc needs to be cleaned up - both bitbucket and dropbox currently modelled after Gtihub as placeholders.

Would like to hand this over - can send a git `patch` if preferred to a PR, or else provide admin rights to the tawawa lock repo.


gnandretta:
Thanks @arcseldon. I was not able to use this as a starting point, so merged #385 instead.

PR #386, opened 2016-05-12
Use optimised version of our logo (badge)

ramasilveyra:
Use optimised version of badge image.
From 8kb to 2kb.

### DIFF
--- a/lib/options-manager/index.js
+++ b/lib/options-manager/index.js
@@ -135,7 +135,7 @@ function OptionsManager(widget, options) {
   if (this.icon) {
     this.headerIcon = this.icon;
   } else {
-    this.headerIcon = "//cdn.auth0.com/styleguide/1.0.0/img/badge.png";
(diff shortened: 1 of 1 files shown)


lib/options-manager/index.js
```diff
@@ -135,7 +135,7 @@ function OptionsManager(widget, options) {
   if (this.icon) {
     this.headerIcon = this.icon;
   } else {
-    this.headerIcon = "//cdn.auth0.com/styleguide/1.0.0/img/badge.png";
+    this.headerIcon = "//cdn.auth0.com/styleguide/latest/lib/logos/img/badge.png";
```
gnandretta: @ramasilveyra not sure how is versioning being handled on styleguide but is there a way to depend on something else than `latest`?

ramasilveyra: @gabrielmoreira Yes, I will change to `4.6.1`

gnandretta: @ramasilveyra thanks!

PR #410, opened 2016-05-27
I18n api clean up

gnandretta:


jfromaniello:
should we remove Klingon as well? I haven't seen many applications nor browsers using that.

gnandretta:
@jfromaniello this is for v10 which won't include Klingon, unless you want to submit a PR 👯

Issue #406, opened 2016-05-26
Upgrade to react 15 (lock 10)


juhaelee:
Lock 10 has a dependency on React 0.14. Can we upgrade it to react 15 and move it to peer dependencies as well?

gnandretta:
Hi, we want to solve the compatibility issue before the next release. However, we are not so sure about listing React as a peer dependency.

From [npm docs](https://docs.npmjs.com/files/package.json#peerdependencies).
> npm versions 1 and 2 will automatically install peerDependencies if they are not explicitly depended upon higher in the dependency tree. In the next major version of npm (npm@3), this will no longer be the case. You will receive a warning that the peerDependency is not installed instead.

This will be a problem for people using Lock in an app that doesn't depend on React. Also, it doesn't look like peer dependencies are meant to be used for that (from the same doc page).

> In some cases, you want to express the compatibility of your package with a host tool or library, while not necessarily doing a require of this host. This is usually referred to as a plugin. Notably, your module may be exposing a specific interface, expected and specified by the host documentation.

I think the way to go is to list React as a regular dependency, with version `"^15.0.0 || ^16.0.0"` as the [React team suggested in their blog](https://facebook.github.io/react/blog/2016/02/19/new-versioning-scheme.html#avoiding-the-major-cliff). What do you think? Would that work for you?

juhaelee:
@gnandretta Yes that would work for me!

juhaelee:
@gnandretta Just created a pull request if you could take a look!

gnandretta:
@juhaelee merged, thank you.

juhaelee:
@gnandretta awesome, could you cut a new release? 🙏

gnandretta:
@juhaelee we still need to finish some things that will go in the next release, can you use a git dependency to your fork in the meantime?

juhaelee:
will do, thanks @gnandretta!

PR #418, opened 2016-06-01
Upgrade to react 15

juhaelee:
@gnandretta fixes https://github.com/auth0/lock/issues/406. Upgrades to react 15 rather than move it to peer dependencies as discussed here: https://github.com/auth0/lock/issues/406#issuecomment-222999828


gnandretta:
@juhaelee  thanks!

PR #461, opened 2016-06-27
Place title field inside of languageDictionary object

smithgeek:


gnandretta:
@smithgeek thank you!

PR #471, opened 2016-07-04
Translate to Latvian

markspolakovs:
Translate the lock widget to Latvian.


markspolakovs:
Oops... accidentally deleted en.json

PR #485, opened 2016-07-14
Show userBlockedErrorText when user is blocked

nicosabena:
Fixes how a blocked user is recognized (status 429, `err.code === 'too_many_attempts'`).
Uses `dict.signin.userBlockedErrorText` to display or falls back to err.message if not configured.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1187,7 +1187,9 @@ Auth0Lock.prototype._signinWithAuth0 = function (panel, connection) {
     self._focusError(email_input);
     self._focusError(password_input);
 
-    if (err.status !== 401) {
+    if ('too_many_attempts' === err.code) {
+        self._showError(self.options.i18n.t('signin:userBlockedErrorText') || err.message);
+    } else if (err.status !== 401) {
       self._showError(self.options.i18n.t('signin:serverErrorText'));
     } else if ('password_change_required' === err.code) {
       self._showError(self.options.i18n.t('signin:passwordChangeRequiredErrorText'));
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1321,13 +1323,13 @@ Auth0Lock.prototype._signinPopupNoRedirect = function (connectionName, popupCall
       self._showError(self.options.i18n.t('signin:userConsentFailed'));
     } else if (err.status === 0) {
       self._showError(self.options.i18n.t('networkError'));
-    } else if (err.status !== 401) {
-      self._showError(self.options.i18n.t('signin:serverErrorText'));
-    } else if ('unauthorized' === err.code && err.details && err.details.error_description === 'user is blocked') {
```
gnandretta: @nicosabena if we remove this `if` we will no longer handle the case when the user is blocked from the dashboard

index.js
```diff
@@ -1321,13 +1323,13 @@ Auth0Lock.prototype._signinPopupNoRedirect = function (connectionName, popupCall
...
+    } else if ('too_many_attempts' === err.code) {
       message = self.options.i18n.t('signin:userBlockedErrorText');
-      self._showError(message || err.details.error_description);
+      self._showError(message || err.message);
```
gnandretta: A different key should be used for this error, for reset we use `tooManyRequestsErrorText`.
Also, we should always get the text from the dict (we only allow to display the message returned by the server in some special cases involving rules).

nicosabena:
Oh, ok, I thought that `signin:userBlockedErrorText` was for users blocked from anomaly detection, but I see now that this is meant for users blocked from rules/dashboard.

Implementing a message for the anomaly detection case means adding a new string resource translated to all languages, so let's close this PR and I will start fresh.

PR #487, opened 2016-07-18
Event wrapping

gnandretta:
This is supposed to ease the integration with frameworks that use dirty checking. Just publishing here to have a discussion starting point.


gnandretta:
We decided to handle this outside lock.

PR #522, opened 2016-08-01
XHTML compatibility fixes.

rafalwrzeszcz:
Fixed Lock to work with XHTML applications (`application/xhtml+xml` Content-Type).

### DIFF
--- a/src/i18n/en.js
+++ b/src/i18n/en.js
@@ -93,6 +93,6 @@ export default {
   usernameOrEmailInputPlaceholder: "username/email",
   title: "Auth0",
   welcome: "Welcome %s!",
-  windowsAuthInstructions: "You are connected from your corporate network&#8230;",
+  windowsAuthInstructions: "You are connected from your corporate network&hellip;",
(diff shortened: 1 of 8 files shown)


gnandretta:
@rafalwrzeszcz can you revert the changes from `&hellip;` to `&#8230;`? That way you can override the translations that are not compatible with xhtml and users who don't need that can still use `&hellip;`.

rafalwrzeszcz:
Sure, I can - but `&#<code>;` notation works both in HTML and XHTML/XML, wouldn't it make more sense?

rafalwrzeszcz:
Also - even though this PR makes JS library working with XHTML it's not the only problem. HTML produced by the backend for WS-Fed contains `<input>` tags (no closing `/>`).

I have a "dirty" fix for it locally, but it's not fixable on client-side in proper way, so I don't include it in this PR (I made a wrapper around `Auth0.prototype._renderAndSubmitWSFedForm` to do `String.replace()` on the param before invoking internal function).

gnandretta:
Thank you.

PR #548, opened 2016-08-10
Chinese translation of lock

soulbeing:
Please review and deliver this Chinese translation. Really in need.


gnandretta:
@soulbeing thank you!

PR #551, opened 2016-08-10
added Powered By Auth0 text on badge for free accounts

mike-casas:
Added "Powered By Auth0" to Lock  of just the badge for Free accounts. 
gif lock

### DIFF
--- a/css/index.styl
+++ b/css/index.styl
@@ -24,7 +24,7 @@
     -webkit-box-sizing initial
     -moz-box-sizing initial
     box-sizing initial
-  
+
   svg
     background-color transparent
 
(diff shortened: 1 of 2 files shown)


css/index.styl
```diff
@@ -779,6 +779,11 @@
 
     .auth0-lock-badge
       display inline-block
+      color rgba(white,.7)
+      font-size 14px
+      font-family "Avenir"
```
gnandretta: @beneliflo can you review the this? do we need fallback font families?

mike-casas: that line font-family I think is not necessary we can delete and use the "Avenir Next" of the .auth0-lock.auth0-lock class.

vctrfrnndz:
@gnandretta That should be "Avenir Next"

woloski:
I'd like to change `Powered by` with `Protected with `

robinvdvleuten:
It would be nice if this can be translated!

PR #560, opened 2016-08-15
Added Norwegian bokmål (nb) translation.

strindhaug:
Note: Norwegian bokmål is also (incorrectly) known by the language code `no`, so ideally `no` should redirect to `nb`.


strindhaug:
No idea why the build failed, but the translations works when I copy/paste the translation into the `languageDictionary` option in our project.

strindhaug:
Is it safe to insert `&shy;` entities to indicate hyphenation points in all the translation strings?

gnandretta:
Thanks!

You can use `&shy;` for keys that end with `Instructions`, like `socialLoginInstructions`, but we have been using explicit `<br />` for those. I guess it would make sense to use `&shy;` in error messages but we don't support html there. Feel free to open a feature request for that if you need it.

PR #593, opened 2016-09-05
Added Swedish (sv) translation.

kuljaninemir:


src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
+export default { 
+ error: { 
+   forgotPassword: { 
+     "too_many_requests": "Du har nått gränsen för maximalt antal försök att ändra ditt lösenord. Var vänlig vänta innan du försöker igen.", 
+     "lock.fallback": "Något gick fel när vi försökte ändra ditt lösenord." 
+   }, 
+   login: { 
+     "blocked_user": "Användaren är spärrad.", 
+     "invalid_user_password": "Fel på inloggningsuppgifter.", 
+      "lock.fallback": "Någit gick fel vid inloggningsförsöket.",
```
fiddur: "Något", not "Någit".

And, indentation got bad from this point on.

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+     "lock.fallback": "Något gick fel när vi försökte ändra ditt lösenord." 
+   }, 
+   login: { 
+     "blocked_user": "Användaren är spärrad.", 
+     "invalid_user_password": "Fel på inloggningsuppgifter.", 
+      "lock.fallback": "Någit gick fel vid inloggningsförsöket.", 
+      "lock.invalid_code": "Fel kod.", 
+      "lock.invalid_email_password": "Fel email eller password.", 
+      "lock.invalid_username_password": "Fel användarnamn eller lösenord.", 
+      "lock.network": "Kunde inte kontakta servern. Var vänlig kontrollera din uppkoppling och försök igen.",
```
fiddur: While not directly wrong, in Swedish it's more common to leave out the "please" part in an instruction, so I would suggest changing the last sentence to just "Kontrollera din uppkoppling och försök igen."

fiddur:
Not to be picky, but comparing with the en.js, I don't think this repo uses tabs for indentation, just two spaces :)

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+			"lock.fallback" : "Något gick fel vid inloggningsförsöket.",
+			"lock.invalid_code" : "Fel kod.",
+			"lock.invalid_email_password" : "Fel e-postadress eller lösenord.",
+			"lock.invalid_username_password" : "Fel användarnamn eller lösenord.",
+			"lock.network" : "Kunde inte kontakta servern. Var vänlig kontrollera din uppkoppling och försök igen.",
+			"lock.popup_closed" : "Popup-fönster stängt. Försök igen.",
+			"lock.unauthorized" : "Åtkomst nekad. Försök igen.",
+			"password_change_required" : "Du behöver ändra ditt lösenord då detta är första gången du loggar in eller för att ditt lösenord har gått ut.", // TODO: verify error code
+			"password_leaked" : "Din konto har blivit låst då ditt lösenord har lagts upp på en annan webbsida. Du har fått e-post med instruktioner om hur du låser upp ditt konto.",
+			"too_many_attempts" : "Ditt konto har blivit låst efter att flera felaktiga inloggningsförsök har gjorts."
```
fiddur: At row 8 the word "blocked" is translated to "spärrad".  I think that is better than "låst", but at least use the same translation.  And, I would use a more terse passive form for the latter part: "Ditt konto har blivit spärrat efter flera felaktiga inloggningsförsök."

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+			"password_no_user_info_error" : "Lösenordet baseras på personuppgifter.",
+			"password_strength_error" : "Lösenordet är för svagt.",
+			"user_exists" : "Användaren finns redan.",
+			"username_exists" : "Användarnamnet finns redan."
+		}
+	},
+	success : { // success messages show above the form or in a confirmation pane
+		logIn : "Tack för att du loggar in.",
+		forgotPassword : "Vi har skickat dig ett e-postmeddelande där du kan återställa ditt lösenord.",
+		magicLink : "Vi har skickat en länk för att logga in<br />till %s.",
```
fiddur: "logga in på", not "till".

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+	},
+	blankErrorHint : "Får inte lämnas tom",
+	codeInputPlaceholder : "Din kod",
+	databaseEnterpriseLoginInstructions : "",
+	databaseEnterpriseAlternativeLoginInstructions : "eller",
+	databaseSignUpInstructions : "",
+	databaseAlternativeSignUpInstructions : "eller",
+	emailInputPlaceholder : "dinepostadress@exempel.se",
+	enterpriseLoginIntructions : "Logga in med dina företagsuppgifter.",
+	enterpriseActiveLoginInstructions : "Ange dina företagsuppgifter för %s.",
```
fiddur: I'm not sure of the context here.  The English term says "Please enter your coorporate credentials at %s.".  If "at" in this context indicates a place to enter the credentials, it should be "vid" rather than "för".  Otoh, if "at" indicates the company name, "för" works fine.

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+	failedLabel : "Misslyckades!",
+	forgotPasswordAction : "Glömt ditt lösenord?",
+	forgotPasswordInstructions : "Ange din e-postadress. Vi skickar ett e-postmeddelande med instruktioner för att återställa ditt lösenord.",
+	forgotPasswordSubmitLabel : "Skicka e-post",
+	invalidErrorHint : "Ogiltig",
+	lastLoginInstructions : "Senast du loggade in med",
```
fiddur: I believe this should be "Senast loggade du in som".  I can't be completely sure, since you really never should translate parts of a sentence and there is no place holder here, but indicating which account you used last time, it would read as "Senast loggade du in som foo@bar.baz".  "med" would indicate the actual credentials, and the order of the words here would indicate only the start of a longer sentence, like "Senast du loggade in med dessa uppgifter så gick det bra…"

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+	loginAtLabel : "Logga in på %s",
+	loginLabel : "Logga in",
+	loginSubmitLabel : "Logga in",
+	loginWithLabel : "Logga in med %s",
+	notYourAccountAction : "Inte ditt konto?",
+	passwordInputPlaceholder : "ditt lösenord",
+	passwordStrength : {
+		containsAtLeast : "Måste innehålla minst %d av följande %d typer av tecken:",
```
fiddur: "types" would be better translated as "sorter", in this case with an s-suffix, "…följande %d sorters tecken:".

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+  loginLabel: "Logga in",
+  loginSubmitLabel: "Logga in",
+  loginWithLabel: "Logga in med %s",
+  notYourAccountAction: "Inte ditt konto?",
+  passwordInputPlaceholder: "ditt lösenord",
+  passwordStrength: {
+    containsAtLeast: "Måste innehålla minst %d av följande %d typer av tecken:",
+    identicalChars: "Inte fler än %d identiska tecken i rad (t.ex., \"%s\" är inte tillåtet)",
+    nonEmpty: "Lösenord får inte vara tomma",
+    numbers: "Nummer (t.ex. 0-9)",
```
fiddur: "Siffror" rather than "Nummer".  On the other hand, I'd say that the English is also wrong and should be "Digits".  See https://en.wikipedia.org/wiki/Numerical_digit

In Swedish we make it even more complicated by having "nummer", "siffra", and "tal".

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+  loginWithLabel: "Logga in med %s",
+  notYourAccountAction: "Inte ditt konto?",
+  passwordInputPlaceholder: "ditt lösenord",
+  passwordStrength: {
+    containsAtLeast: "Måste innehålla minst %d av följande %d sorters tecken:",
+    identicalChars: "Inte fler än %d identiska tecken i rad (t.ex., \"%s\" är inte tillåtet)",
+    nonEmpty: "Lösenord får inte vara tomma",
+    numbers: "Nummer (t.ex. 0-9)",
+    lengthAtLeast: "Minst %d tecken långt",
+    lowerCase: "Gemener (a-z)",
```
fiddur: Technical question: Do we only allow those letters, or should this indicate the whole alphabet (a-ö in Swedish)?   I would guess the former, but just to be sure…

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+    shouldContain: "Måste innehålla:",
```
fiddur: The translation goes from "should" to "must", but if it is mandatory, it is a "must" and then "Måste" works well even though it's not a direct translation.

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+    specialCharacters: "Specialtecken (t.ex. !@#$%^&*)",
+    upperCase: "Versaler (A-Z)"
+  },
+  passwordlessEmailAlternativeInstructions: "Annars kan du ange din e-postadress för att logga in<br>eller skapa ett konto",
```
fiddur: Hmm, this has gone from an instruction to an information about a possibility (like "Otherwise you could…").  It's a bit hard to translate without seeing what precedes this, but I think "Annars, ange din e-postadress …" without "kan".

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+  passwordlessEmailCodeInstructions: "Ett e-postmeddelande med din kod har skickats till %s.",
+  passwordlessEmailInstructions: "Ange din e-postadress för att logga in<br>eller skapa ett konto",
+  passwordlessSMSAlternativeInstructions: "Annars kan du ange ditt telefonnummer för att logga in<br>eller skapa ett konto",
```
fiddur: Same thing here, "Annars, ange ditt…"

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+  passwordlessSMSCodeInstructions: "Ett SMS med din kod har skickats<br>till %s.",
```
fiddur: "the code" would be "koden", not "din kod"

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+  passwordlessSMSInstructions: "Ange ditt telefonnummer för att logga in<br>eller skapa ett konto",
+  phoneNumberInputPlaceholder: "ditt telefonnummer",
+  resendCodeAction: "Fick du ingen kod?",
+  resendLabel: "Skicka om",
+  resendingLabel: "Skickar om...",
```
fiddur: "Skicka igen" rather than "om"

src/i18n/sv.js
```diff
@@ -0,0 +1,102 @@
...
+  signUpLabel: "Registrera dig",
+  signUpSubmitLabel: "Registrera dig",
+  signUpTerms: "",
+  signUpWithLabel: "Registrera dig med %s",
+  socialLoginInstructions: "",
+  socialSignUpInstructions: "",
+  ssoEnabled: "Single Sign-On aktiverad",
+  submitLabel: "Skicka",
+  unrecoverableError: "Något gick fel.<br />Vänligen kontakta teknisk support.",
+  usernameFormatErrorHint: "Använd 1-15 tecken, siffror och \"_\"",
```
fiddur: "letters" would be "bokstäver", not "tecken".

kuljaninemir:
@fiddur A lot of good points.  I'm not the one that actually translated this, but I should have read through it before doing a pull request.

fiddur:
It's not really bad, and the user would understand, but I was asked to review it and I might be a bit of a perfectionist :)

src/i18n/sv.js
kuljaninemir: I think this refers to the authentication connection, like Google or Facebook. "Senast loggade du in med Google" makes sense. I'm not entirely sure that it can't refer to user credentials though.

PR #595, opened 2016-09-06
use ReactCSSTransitionGroup for global messages

robbiewxyz:
`global_message.jsx`'s code for componentWillEnter and componentWillLeave looked like quite a mess with the .style.this and .style.that.  This change lets the CSS do what it's good at and keeps the JS looking a bit cleaner in that file.

It also cuts down on a dependency.

### DIFF
--- a/css/index.styl
+++ b/css/index.styl
@@ -47,6 +47,22 @@ iphoneBar = 68px
     &.auth0-global-message-success
       background #7ED321
 
+    &.global-message-enter
+      height 0
+      paddingTop 0
+      paddingBottom 0
+    &.global-message-enter.example-enter-active
+      transition all 0.2s
+      height auto
+      paddingTop 10px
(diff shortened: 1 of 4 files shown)


css/index.styl
```diff
@@ -47,6 +47,22 @@ iphoneBar = 68px
     &.auth0-global-message-success
       background #7ED321
 
+    &.global-message-enter
+      height 0
+      paddingTop 0
+      paddingBottom 0
+    &.global-message-enter.example-enter-active
```
cristiandouce: What is `.example-enter-active` for?

robbiewxyz: D'oh.  That's from my testing; I forgot to replace it out.  It's fixed now.

robbiewxyz:
Hey, just curious:  what is your release cycle like?  When will a pull like this shows up in the NPM?

PR #605, opened 2016-09-13
Remove absolute positioning that brakes alignment on small screens

aouji:
The absolute positioning introduced in https://github.com/auth0/lock/pull/597 breakes alignment as the siblings of the tabs-container do not get absolute positioning in small screens. Also since we have to go high up in the parents hierarchy to get a positioned parent, it should be better to not have the absolute positioning here. I also checked that the terms still remain beneath password:

Before:
screen shot 2016-09-12 at 5 55 48 pm

After:
screen shot 2016-09-12 at 5 50 20 pm

### DIFF
--- a/css/index.styl
+++ b/css/index.styl
@@ -924,15 +924,6 @@ tabsHeight = 40px
       .auth0-lock-form p
         @media screen and (max-width: 480px)
           font-size 14px
-    
-    .auth0-lock-tabs-container
-      //mobile view
-      @media screen and (max-width: 480px)
-        position absolute
-        top 0
-        left 0
-        width 100%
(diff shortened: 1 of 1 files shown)


aouji:
@cristiandouce Sorry to ping, but it will be great if I can get an opinion on this. I am currently migrating our auth0 applications from lock 7 to lock 10.2 and without this (or some other fix) I have to manually override the css in our lock pages.

hzalaz:
@aouji we fixed in master and we'll probably release it next week. Will keep this open till we release

hzalaz:
Closing this @aouji please let us know if it's still an issue

PR #628, opened 2016-09-29
MFA when using  oauth/ro endpoint

dafortune:
This enables lock to use MFA for ajax calls to the RO endpoint.
The basic configuration to support this use case is:

```
    const lock = new Auth0Lock(cid, domain, {
      auth: {
        redirect: false,
        sso: false
      }
    });
    lock.show();
    lock.on("authenticated", function(authResult) {
      lock.hide();
      alert(JSON.stringify(authResult, null, 2));
    });
```

The result is that in case of mfa required we will show a new screen:
screen shot 2016-09-29 at 12 29 28 am


src/i18n/en.js
```diff
@@ -14,6 +14,8 @@ export default {
       "lock.network": "We could not reach the server. Please check your connection and try again.",
       "lock.popup_closed": "Popup window closed. Try again.",
       "lock.unauthorized": "Permissions were not granted. Try again.",
+      "lock.mfa_registration_required": "Multifactor authentication is required but your device is not enrolled. Please enroll it before moving on.",
```
dafortune: Not sure what to say in this case; we don't support enrollment. "Enrollment not supported"?

narwajea:
Why just update src/i18n/en.js & src/i18n/es.js ? I get many warnings in fr locale..

glena:
@narwajea you can add disableWarnings:true to disable them

narwajea:
Thank you for the suggestion, but still why not updating all language files ?

glena:
Yes we are working on it, but there are a lot of languages and we are trying to avoid using google translate to assure the quality of the translation. If you are fluent in french and want to collaborate PR are always welcome.

PR #642, opened 2016-10-13
Single AD connection without domain shows undefined in message

### LINKED ISSUE #627: Single AD connection without domain shows undefined in message
When the client only allows a single AD connection, Lock will display the form for username/password with the title `Please enter your coorporate credentials at {domain}`. We need to avoid showing the `undefined` by changing the message or hiding for this specific case

glena:
Closes #627

### DIFF
--- a/src/connection/enterprise/hrd_screen.jsx
+++ b/src/connection/enterprise/hrd_screen.jsx
@@ -11,7 +11,7 @@ const Component = ({i18n, model}) => {
 
   var headerText;
   
-  if (domain) {
+  if (domain != null) {
     headerText = i18n.html("enterpriseActiveLoginInstructions", domain);     
   } else {
     headerText = i18n.html("enterpriseLoginIntructions");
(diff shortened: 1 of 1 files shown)


src/connection/enterprise/hrd_screen.jsx
```diff
@@ -7,7 +7,18 @@ import { enterpriseDomain, isSingleHRDConnection }  from '../enterprise';
 import * as i18n from '../../i18n';
 
 const Component = ({i18n, model}) => {
-  const headerText = i18n.html("enterpriseActiveLoginInstructions", enterpriseDomain(model)) || null;
+  const domain = enterpriseDomain(model);
+
+  var headerText;
+  
+  if (domain) {
```
hzalaz: Please change it to 
```js
if (domain != null) {
```

glena: done

PR #646, opened 2016-10-17
Only require non-empty value for username when DB connection is custom or import is enabled

glena:


src/field/username.js
```diff
@@ -8,7 +8,8 @@ const DEFAULT_CONNECTION_VALIDATION = { username: { min: 1, max: 15 } };
 const regExp = /^[a-zA-Z0-9_]+$/;
 
 function validateUsername(str, validateFormat, settings = DEFAULT_CONNECTION_VALIDATION.username) {
-  if (!validateFormat) {
+
+  if (!validateFormat  || !settings) {
```
glena: If the connection does not have validation settings, it should only check if the field is empty.
`validateFormat` overrides this logic to disable validation on login (login should never validate format)

src/field/username/username_pane.jsx
```diff
@@ -38,7 +38,7 @@ export default class UsernamePane extends React.Component {
     const invalidHint = (str) => {
       const hintKey = invalidHintKey(str);
 
-      if ("usernameFormatErrorHint" === hintKey && validateFormat) {
+      if ("usernameFormatErrorHint" === hintKey && validateFormat && usernameValidation) {
```
glena: only show format info in the error, if there is any format restriction.

PR #647, opened 2016-10-17
provide separate translation tokens for email and password hints in i…

elger:
Hi,

in order to be able to set different user feedback information on empty or invalid fields, I added some more i18n tokens.

### DIFF
--- a/src/field/email/email_pane.jsx
+++ b/src/field/email/email_pane.jsx
@@ -31,7 +31,7 @@ export default class EmailPane extends React.Component {
     return (
       <EmailInput
         value={value}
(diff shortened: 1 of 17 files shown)


hzalaz:
Thanks for the PR, but the changes is sadly a breaking one since people with customizations on `invalidErrorHint` and `blankErrorHint` will see a different error (probably in a different language).

The easy fix is to make these values optional in the translation file and in the code default to the old ones if the specific error for the field is not found.

hzalaz:
Closing this one since it can potentially break language strings customizations, if you can avoid breaking existing users like we suggested in a previous comment please reopen the PR and we might evaluate merging it.

PR #648, opened 2016-10-17
Added support for custom oauth2 connections

glena:
```js
const options = { 
      theme: {
        authButtons: {
          "connectionName": {
            displayName: "...", 
            primaryColor: "...", 
            foregroundColor: "...", 
            icon: "url(...)"
          }
        }
      }
    };

const lock = new Auth0Lock(cid, domain, options);
```


src/ui/button/auth_button.jsx
```diff
@@ -27,7 +41,9 @@ AuthButton.propTypes = {
   isBig: React.PropTypes.bool.isRequired,
   label: React.PropTypes.string.isRequired,
   onClick: React.PropTypes.func.isRequired,
-  strategy: React.PropTypes.string.isRequired
+  strategy: React.PropTypes.string.isRequired,
+  icon: React.PropTypes.string,
+  background: React.PropTypes.string
```
cristiandouce: shouldn't this one be `style`?

PR #650, opened 2016-10-18
Fix Issues with Overlay mode + signUpLink setting on a SPA

### LINKED ISSUE #619: Issues with Overlay mode + signUpLink setting on a SPA
This happens when:

- Lock 10 is being used in a SPA
- Lock 10 is being shown as overlay
- `signUpLink` was set for redirecting the user to a custom signup page

When the user clicks on `sign up`, Lock changes the route but since the SPA router avoids doing the actual redirect, lock is still visible on top of the page content.

It should close itself when it redirects the user

glena:
Closes #619

### DIFF
--- a/src/connection/database/login_sign_up_tabs.jsx
+++ b/src/connection/database/login_sign_up_tabs.jsx
@@ -2,9 +2,7 @@ import React from 'react';
 import { showLoginActivity, showSignUpActivity } from './actions';
 import * as l from '../../core/index';
 import { getScreen } from './index';
-import {
-  closeLock
-} from '../../core/actions';
+import { closeLock } from '../../core/actions';
 
 export default class LoginSignUpTabs extends React.Component {
 
--- a/src/connection/database/login_sign_up_tabs.jsx
+++ b/src/connection/database/login_sign_up_tabs.jsx
@@ -2,6 +2,9 @@ import React from 'react';
 import { showLoginActivity, showSignUpActivity } from './actions';
 import * as l from '../../core/index';
 import { getScreen } from './index';
+import {
+  closeLock
+} from '../../core/actions';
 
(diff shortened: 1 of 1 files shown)


src/connection/database/login_sign_up_tabs.jsx
```diff
@@ -2,6 +2,9 @@ import React from 'react';
 import { showLoginActivity, showSignUpActivity } from './actions';
 import * as l from '../../core/index';
 import { getScreen } from './index';
+import {
+  closeLock
```
hzalaz: Not sure what is the style but maybe follow the other import single line style.

hzalaz: @cristiandouce ?

src/connection/database/login_sign_up_tabs.jsx
```diff
@@ -38,9 +42,16 @@ export default class LoginSignUpTabs extends React.Component {
   }
 
   handleSignUpClick() {
+    if (this.props.signUpLink) {
```
hzalaz: Should we consider the `closable` property value with this or should we just ignore it?

src/connection/database/login_sign_up_tabs.jsx
glena: right

src/connection/database/login_sign_up_tabs.jsx
glena: I don't think so, in this context there are 2 scenarios:

regular webapp: the redirection will change the entire page. It is irrelevant if lock hides or not
spa: changing the route will not force a redirection so lock keeps visible unless it is explicitly hidden (this change).

While implementing lock, you might want it to not be closable (forcing signup or login on a page?) but you will want it to hide if the user is redirected to a different signup page.

PR #673, opened 2016-10-25
fixed issue with focus in email and user input

beneliflo:
before:
issue

after:


glena:
Closed in favor of #681

PR #680, opened 2016-10-26
allow to override language, dict, logo and primary color on show method

glena:


vasyabigi:
@glena is it possible to override `container` too? In SPAs it would be convenient to use *Lock* as a modal or inline depending on the route. Is it possible to do today? If not, let's add this possibility to `show` method too.

glena:
@vasyabigi No, it is not posible. You will need to re-initialize lock to change this property.

PR #693, opened 2016-11-01
Support for Turkish language

saltukalakus:
Translation file for Turkish is added.


glena:
Thank you for the contribution

PR #714, opened 2016-11-10
Enterprise: force username for AD connections

### LINKED ISSUE #712: AD/LDAP connection + social asks for email instead of username
This is Lock v9 for a client that has only an AD/LDAP connection enabled on the left, and the same client with an AD/LDAP connection + a social connection on the right. Things look good:

image

In Lock v10, however, same setup:

image

Instead of asking for username/password, it asks for an email/password.

glena:
Closes #712

### DIFF
--- a/src/connection/database/index.js
+++ b/src/connection/database/index.js
@@ -278,9 +278,13 @@ export function databaseConnectionRequiresUsername(m) {
 }
 
 export function databaseUsernameStyle(m) {
-  return databaseConnectionRequiresUsername(m)
-    ? get(m, "usernameStyle", "any")
-    : "email";
+  if (l.hasSomeConnections(m, "database")) {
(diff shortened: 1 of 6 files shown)


glena:
One enterprise connection with domain
image

One enterprise connection without domain
image

Two enterprise connections with and without domain
image

src/connection/database/index.js
```diff
@@ -10,6 +10,9 @@ import { dataFns } from '../../utils/data_utils';
 import sync from '../../sync';
 import trim from 'trim';
 import { defaultDirectory } from '../../core/tenant';
+import {
+  findADConnectionWithoutDomain
```
hzalaz: Not sure what is our style but the codebase has single line imports 😄

PR #719, opened 2016-11-12
Correcting callbackUrl to be redirectUrl

leog:
According to the Lock 9 to Lock 10 migration doc the `callbackUrl` was renamed to `redirectUrl`.


hzalaz:
Thanks for the PR!

johnlim:
Should/will `redirectUrl` be renamed to `redirectUri` so that its consistent with auth0.js V8?

luisrudge:
That would break a lot of users, so maybe in the next major version.

johnlim:
@luisrudge thanks! looking forward to it.

Issue #741, opened 2016-12-02
Double Tabbing necessary when filling out username/password in IE11 with keyboard only


MaximRouiller:
**Version:** Lock10
**Issue:** When tabbing between username and password, SVG gets focused
**Browser:** IE11 only

**Workaround:**

```javascript
var lock = new Auth0Lock(appConfig.clientID, appConfig.domain, options);
lock.on('signin ready', function () {
    $('svg').attr('focusable', false);
});
```
**Root cause:** SVG elements can get focus when tabbing in IE11.

hzalaz:
Should be fixed in latest Lock

PR #742, opened 2016-12-02
Adding focusable='false' to password SVG element

MaximRouiller:
Solves #741.

Might be necessary to review ALL svg.


hzalaz:
@MaximRouiller thanks for the PR and the related issue. We'll send another PR setting focusable to all SVG just in case.

MrPrashantT:
@MaximRouiller @hzalaz  which version of Lock has this been merged into?

luisrudge:
@MrPrashantT here: https://github.com/auth0/lock/pull/873 - this will be in the next release

MaximRouiller:
@MrPrashantT 

In the mid time, here are your options:

* you can either modify the source directly (do not recommend)
* you can do a fork on GitHub, apply the patch and use this (urghhh...)
* Monkey patch! Using jQuery, select all svg and add the attribute `focusable="false"` to all of them.

Most of the time, you won't want your SVG to have focus anyway. It's only IE who's a dirty dog in doing this. Let me know if I can help you further.

luisrudge:
We just released a new version of lock with this fix. Thank you all for the patience!

PR #743, opened 2016-12-03
Shorten Russian signUpLabel to fit into the widget width

cyxou:
Please change the Russian translation for the signUpLabel key from "Зарегистрироваться" to "Регистрация". Semantically it's the same but "Зарегистрироваться" is longer and does not fit well into the width of container. Consider the following screenshots:

image

image


cyxou:
Thanks a lot =)

glena:
Thank you!

PR #745, opened 2016-12-06
Hide first screen title option

glena:
Help defining the option name, now is `hideFistPageTitle`


glena:
@beneliflo we need to review the css

hzalaz:
@glena not for now but we need to think the flag name and rebase 😄

glena:
`hideMainScreenTitle` it is

PR #749, opened 2016-12-14
Change the tenant info endpoints

glena:


src/core/index.js
```diff
@@ -265,7 +265,7 @@ function extractClientBaseUrlOption(opts, domain) {
 
 export function extractTenantBaseUrlOption(opts, domain) {
   if (opts.configurationBaseUrl && typeof opts.configurationBaseUrl === "string") {
-    return opts.configurationBaseUrl;
+    return urljoin(opts.configurationBaseUrl, 'info-v1.js');
```
glena: now `configurationBaseUrl` is only the domain, it will append the blob endpoint.

src/core/index.js
```diff
@@ -285,11 +285,12 @@ export function extractTenantBaseUrlOption(opts, domain) {
     domain = parts.length > 3
       ? "https://cdn." + parts[parts.length - 3] + DOT_AUTH0_DOT_COM
       : AUTH0_US_CDN_URL;
+
+    return urljoin(domain, 'tenants', 'v1', `${tenant_name}.js`);
```
glena: for cloud accounts we are going to get the blob like before from the CDN.

PR #752, opened 2016-12-18
Create fa.js

doroudi:
Persian Language Library for Auth0 Lock


glena:
thank you for the contribution

Issue #543, opened 2016-08-08
[v10] Enterprise connections don't strip domain from email


nicosabena:
Say you have an enterprise connection, with `fabrikam.com` set as domain (for HRD). If the user types `jon@fabrikam.com`, he will be asked to type his corporate credentials (both Lock versions will even suggest `jon` as the username).

**If, at this moment, the user types his or her email** ​_again_​ (thinking that this is the username Lock is expecting), :)
- In Lock v9 the code automatically strips the domain (`@fabrikam.com`) (see https://github.com/auth0/lock/blob/v9/index.js#L1144). So, since only the username is sent in the background, authentication succeeds anyway. (By the way, this only works for the first domain of an enterprise connection).
- In Lock v10, the whole email is sent as the username to the authentication endpoint, so the user gets an "Invalid username or password" error.

Is this difference on purpose and users should type the correct username now, or was this an omission and Lock v10 should behave like Lock v9 for this?

PR #784, opened 2017-01-09
Don't emit error when registering for event 'signin ready'

theopak:
### Problem 

Looks like Lock v10.8.0 added event validation via https://github.com/auth0/lock/pull/756. However this feature emits error messages for many events that Lock makes available (e.g. those events from https://github.com/auth0/docs/pull/191). 

This bug occurs on all supported browsers and operating systems since at least the v10.8.0 release, including the latest version today.

### Fix

I simply added 'signin ready' to the whitelist of valid events: https://github.com/auth0/lock/blob/master/src/core.js#L36

There are other events in https://github.com/auth0/docs/pull/191 that have the same issue but for my purposes I only need 'signin ready' fixed in order to unblock my production issue. 

When I build this branch locally I no longer see the 'Invalid event' console error.

### Issue Reproduction

The following code in my application results in the error `Invalid event "signin ready".`.

Here's the same code in a codepen so you can confirm that 'Invalid event' shows up in the browser console: http://codepen.io/theopak/pen/QdjVZX

```js
lock.show()

// This function adds a message to the Lock.js dialog box, 
// as a proof-of-concept of something that can only be done
// after the 'signin ready' event is triggered.
const addMessageToLock = (event) => {
  let myDiv = document.createElement('h4')
  myDiv.style.textAlign = 'center'
  myDiv.style.textAlign = 'center'
  myDiv.innerHTML = 'Example Message Added to Lock'
  const targetNode = document.getElementsByClassName('auth0-lock-input-block')[0].parentNode
  targetNode.insertBefore(myDiv, targetNode.firstChild)
}

// Here's an example of how the 'signin ready' event is used
// to do something that's only possible after Lock.js is in 
// the DOM and done appearing.
lock.on('signin ready', addMessageToLock)
```

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -39,7 +39,8 @@ export default class Base extends EventEmitter {
       'unrecoverable_error',
       'authenticated',
       'authorization_error',
-      'hash_parsed'
+      'hash_parsed',
+      'signin ready'
     ];
 
     this.id = idu.incremental();
--- a/src/core.js
+++ b/src/core.js
@@ -40,7 +40,8 @@ export default class Base extends EventEmitter {
       'authenticated',
       'authorization_error',
       'hash_parsed',
-      'signin ready'
+      'signin ready',
+      'signup ready'
     ];
 
     this.id = idu.incremental();


theopak:
Hey Auth0 team, it would be much appreciated if you could consider this minor fix to be included on both a v10.9.x and v10.8.x release ASAP because it's causing some annoyance among teams at our company that have become dependent on Lock's 'signin ready' event in order to apply a specific customization that we need. Thanks!

src/core.js
```diff
@@ -39,7 +39,8 @@ export default class Base extends EventEmitter {
       'unrecoverable_error',
       'authenticated',
       'authorization_error',
-      'hash_parsed'
+      'hash_parsed',
+      'signin ready'
```
glena: can you add `signup ready` too?

glena:
This will impact lock 10.9. I think the safer option for you is to use the previous release to the one that is `10.7.2`.

src/core.js
theopak: sure, done

hzalaz:
Also this is sort of an undocumented event, while I understand the use case and need, we will review it to see if it works properly since we didn't intended it to be publicly used.

Once we make sure it's working properly and documented, we'll release it for v10.9.x

theopak:
> I think the safer option for you is to use the previous release to the one that is 10.7.2.

That works too, although then we'd have other issues that were already fixed. v10.8.0 fixing #543 was especially nice!

> This will impact lock 10.9.

Do you mean that, if merged, this could be included in a v10.9.x release? Sounds good to me.

theopak:
> Also this is sort of an undocumented event, while I understand the use case and need, we will review it to see if it works properly since we didn't intended it to be publicly used.

@hzalaz Thanks! As background information, I first found this event from a documentation page under https://auth0.com/docs/libraries/lock. IIRC that doc was lost when Lock v10 came out although you can see the 'signin ready' event advertised in an old commit of the page: https://github.com/auth0/docs/blob/5746f715c00785e47a5eb5e9d47fc41e95b45451/articles/libraries/lock/events.md

glena:
Yes, that is the documentation of lock < 10

theopak:
Hi @glena I believe I made the change you requested. Could you review this PR again when you have a minute?

glena:
released in `10.9.2`

PR #790, opened 2017-01-10
Add resumeAuth method and autoParseHash flag

luisrudge:
When Lock redirects to the redirect url,  it puts the params after a `#` (eg: `#id_token=foo&access_token=bar`. This causes some issues with client side routers that use `#` to handle urls. An example is https://github.com/auth0-samples/auth0-angularjs2-systemjs-sample/issues/40.

This PR adds an option to prevent parsing the hash when you instantiate lock (`autoParseHash`) and adds the `resumeAuth` method to resume the authentication steps  (including all the events)

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -4,6 +4,7 @@ import { remove, render } from './ui/box';
 import webAPI from './core/web_api';
 import {
   closeLock,
+  resumeAuth,
   handleAuthCallback,
   openLock,
   removeLock,
@@ -50,14 +51,13 @@ export default class Base extends EventEmitter {
     go(this.id);
 
     let m = setupLock(this.id, clientID, domain, options, hookRunner, emitEventFn);
-
     this.on('newListener', (type) => {
       if (this.validEvents.indexOf(type) === -1) {
         l.emitUnrecoverableErrorEvent(m, `Invalid event "${type}".`)
       }
     });
 
-    if (!Base.hasScheduledAuthCallback) {
+    if (l.auth.autoParseHash(m) && !Base.hasScheduledAuthCallback) {
       Base.hasScheduledAuthCallback = true;
       setTimeout(handleAuthCallback, 0);
     }
@@ -132,6 +132,10 @@ export default class Base extends EventEmitter {
         remove(l.ui.containerID(m));
       }
     });
+  }
+
+  resumeAuth(hash, callback) {
+    resumeAuth(hash, callback);
   }
 
   show(opts = {}) {
--- a/README.md
+++ b/README.md
@@ -126,6 +126,22 @@ lock.show();
 lock.show({allowedConnections: ["twitter", "facebook"]})
 ```
 
+### resumeAuth(hash, callback)
+
+If you set the [auth.autoParseHash](#authentication-options) option to `false`, you'll need to call this method to complete the authentication flow. This method is useful when you're using a client-side router that uses a `#` to handle urls (angular2 with `useHash` or react-router with `hashHistory`).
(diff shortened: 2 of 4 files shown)


src/core.js
```diff
@@ -50,16 +51,17 @@ export default class Base extends EventEmitter {
...
       if (this.validEvents.indexOf(type) === -1) {
         l.emitUnrecoverableErrorEvent(m, `Invalid event "${type}".`)
       }
     });
 
-    if (!Base.hasScheduledAuthCallback) {
-      Base.hasScheduledAuthCallback = true;
-      setTimeout(handleAuthCallback, 0);
+    if (!options.disableParseHashOnInitialization) {
+      if (!Base.hasScheduledAuthCallback) {
```
hzalaz: Just merge the ifs there is no need to nest them

stefanruijsenaars:
Are there any code snippets that show how can incorporate this into an Angular 2 application with useHash = true?

luisrudge:
@stefanruijsenaars I think @chenkie is updating our angular2 guides. For now, use [this](https://github.com/auth0-samples/auth0-angularjs2-systemjs-sample/issues/40#issuecomment-265170465) as a guide. Then replace `handleRedirectWithHash` with this:

```js
private handleRedirectWithHash() {
  this.router.events.take(1).subscribe(event => {
    if (/access_token/.test(event.url) || /error/.test(event.url)) {
      this.lock.resumeAuth(window.location.hash);
    }
  });
}
```

I'm not sure this will work 100%, but it's something like that. Our angular2 guide will work 100% though, if you want to wait for it 😁

chenkie:
@stefanruijsenaars the method that's going into the samples will pretty much look like this:

```ts
public handleAuthenticationWithHash(): void {
  this
    .router
    .events
    .filter(event => event.constructor.name === 'NavigationStart')
    .filter(event => (/access_token|id_token|error/).test(event.url))
    .subscribe(event => {
      this.lock.resumeAuth(window.location.hash, (error, authResult) => {
        if (error) return console.log(error);
        this.setUser(authResult);
        this.router.navigateByUrl('/');
      });
  });
}

private setUser(authResult): void {
  localStorage.setItem('access_token', authResult.accessToken);
  localStorage.setItem('id_token', authResult.idToken);
}
```

mrahbar:
@chenkie For someone how tries to get auth0 working in angular 1 as well as 2 I find unfortunately lot of different examples and configuration options and can't always tell which is the correct and most up to date one. 

Having said that, your code snippet should be take as the **current** way to implement auth0-lock on angular 2. The method handleAuthCallback() path didn't work as expected because it was called after location.hash was empty. Probably because the problems mentioned here  
- https://github.com/auth0-samples/auth0-angularjs2-systemjs-sample/issues/40#issuecomment-265170465 
- https://github.com/auth0/lock/issues/806

Now resumeAuth() which was called anyways by handleAuthCallback() has to be called directly via the above provided implementation. Which in turn will lead to the emitting of the existing events called in parseHash() callback.

#### So my two cents:
I kept my previous implementation of ` this.lock.on("authenticated", (authResult) => { ...} )` ( and added also an event listener on 'authorization_error' just in case) and used the following snippet which is based on the above code:

```
 this._router.events
      .filter(event => event instanceof NavigationStart)
      .filter(event => (/access_token|id_token|error/).test(event.url))
      .subscribe(event => {
        this.lock.resumeAuth(window.location.hash, (error, authResult) => {});
      });
```

This lets existing code the way it is as well as existing examples and tutorial snippets. And remember to set auth: { autoParseHash: false} in the config options.

PR #805, opened 2017-01-13
Kerberos network checking no longer depends on rememberLastLogin

### LINKED ISSUE #852: Bring back the integratedWindowsLogin option?
Lock v9 had an `integratedWindowsLogin` option which defaulted to `true` that allowed skipping the blue "Windows Authentication" button for Kerberos signin when set to `false` (code [here](https://github.com/auth0/lock/blob/07d0e45444e0853afa35b76b0a9648278d56e9cd/index.js#L659-L660)). This option is no longer present in v10 (although [it was documented](https://github.com/auth0/docs/blob/7a07fbe67968a55bee0f67e5ddf2ed705d3ab046/articles/libraries/lock/v10/customization.md#integratedwindowslogin-boolean)).

Lock currently shows that button based on the IP address (AFAIK), but some users might be in the network and using a Windows machine (or just not connected to AD), so the "Windows Authentication" button doesn't always make sense.

Not sure if the showing/hiding could be automated somehow, but at least giving the flexibility of setting the `integratedWindowsLogin` option to `false` seems like a handy alternative.

Note that `rememberLastLogin: false` can be used now to skip the Windows login button, but this may stop working if [this PR](https://github.com/auth0/lock/pull/805) gets merged.

patrickmcgraw:
Closes #852


patrickmcgraw:
BTW, this resulted from help desk ticket 18786. Nicolás Sabena asked me to open this PR.

hzalaz:
Thanks @patrickmcgraw will review it and let you know

PR #823, opened 2017-01-19
Adds tests to all "field" components 

luisrudge:


src/field/phone-number/phone_number_pane.jsx
```diff
@@ -29,13 +28,13 @@ export default class PhoneNumberPane extends React.Component {
...
         <SelectInput
           icon={icon}
-          isValid={!isFieldVisiblyInvalid(lock, "location")}
+          isValid={!c.isFieldVisiblyInvalid(lock, "location")}
           name="location"
           placeholder=""
           label={humanLocation(lock)}
           onClick={() => startOptionSelection(l.id(lock), "location", "", icon)}
         />
-        <PhoneNumberInput ref="phoneNumberInput"
```
luisrudge: @glena @hzalaz do you think it's safe to the remove this ref? I didn't find anything using it, but maybe it's something more obscure.. What do you think?

glena: that part belongs to the passwordless widget. Probable it is not being used as  you said but couldnt tell. I would left it there and add a `//TODO` so when we start working on that widget, we can assess and test what happens with that. makes sense?

luisrudge: Thanks @glena. @hzalaz said the exact same thing. It's from passwordless and we shouldn't care about it 😁

PR #835, opened 2017-01-27
inject cordova plugin and force popup/sso in cordova or electron

glena:
Should change auth0.js version

### DIFF
--- a/src/core/web_api/p2_api.js
+++ b/src/core/web_api/p2_api.js
@@ -1,4 +1,5 @@
 import auth0 from 'auth0-js';
+import CordovaAuth0Plugin from 'auth0-js/plugins/cordova';
 import * as l from '../index';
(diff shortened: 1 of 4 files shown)


src/core/web_api/p2_api.js
```diff
@@ -1,4 +1,5 @@
 import auth0 from 'auth0-js';
+import CordovaAuth0Plugin from 'auth0-js/plugins/cordova';
```
hzalaz: do we need to add this for p2?

glena: social connections wont work if we do not use the plugin there.

ihsansatriawan:
Hi @glena @hzalaz please check my issue https://github.com/auth0/lock/issues/867 about cordova

Thx

PR #855, opened 2017-02-03
Only set prefill values when client is initialized

### LINKED ISSUE #854: prefill option gives "Please use your corporate email to login"
The prefill option appears to no longer work correctly. The prefill email is correctly filled out in the Lock dialog, but when the user then presses "tab", enter password, "return", a validation error is shown for the email field with the text "Please use your corporate email to login", see the attached screenshot. If the user then clicks the email field, deletes and retypes a single character, and then click "return" then validation succeeds and the form is submitted.

The issue can be reproduced with the HTML code below. Note that the clientId / domain combination that I use is configured for "Username-Password-Authentication" connection.

Browser: Seen in Firefox and Chrome
OS: Seen on Windows

    <script src="http://cdn.auth0.com/js/lock/10.11.0/lock.min.js"></script>
    <script>
    var clientId = '(my clientId)';
    var domain = '(my domain)';
    var options = {
      allowedConnections: [
        'Username-Password-Authentication'
      ],
      prefill: {
        email: 'lars@example.com'
      }
    };
    var lock = new Auth0Lock(clientId, domain, options);
    lock.show();
    </script>
2017-02-02 21_49_34-mozilla firefox

luisrudge:
fix #854

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -38,11 +38,12 @@ export function setup(id, clientID, domain, options, hookRunner, emitEventFn) {
     hashCleanup: options.hashCleanup === false ? false : true,
     allowedConnections: Immutable.fromJS(options.allowedConnections || []),
     ui: extractUIOptions(id, options),
-    defaultADUsernameFromEmailPrefix: options.defaultADUsernameFromEmailPrefix === false ? false : true
+    defaultADUsernameFromEmailPrefix: options.defaultADUsernameFromEmailPrefix === false ? false : true,
+    prefill: Immutable.fromJS(options.prefill || {})
   }));
 
   m = i18n.initI18n(m);
-
+  
   return m;
 }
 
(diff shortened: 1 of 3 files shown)


src/core/index.js
```diff
@@ -38,11 +38,12 @@ export function setup(id, clientID, domain, options, hookRunner, emitEventFn) {
     hashCleanup: options.hashCleanup === false ? false : true,
     allowedConnections: Immutable.fromJS(options.allowedConnections || []),
     ui: extractUIOptions(id, options),
-    defaultADUsernameFromEmailPrefix: options.defaultADUsernameFromEmailPrefix === false ? false : true
+    defaultADUsernameFromEmailPrefix: options.defaultADUsernameFromEmailPrefix === false ? false : true,
+    prefill: Immutable.fromJS(options.prefill || {})
```
glena: Maybe dont wrap it? if you will end up unwrapping if

luisrudge: fixed

PR #860, opened 2017-02-07
Added checkbox CustomInput for additionalSignUpFields

dariobanfi:
Currently additionalSignUpFields parameter lets you use only the type "input" and "select".

In this PR I introduce the CheckBox custom input, which is often very useful in the registration process to accept subscription to the newsletter.

I kept the UI and font similar to the acceptTerms checkbox, but removed the gray background, see here:
screen shot 2017-02-07 at 11 35 30


README.md
```diff
@@ -368,6 +368,23 @@ var options = {
...
+
+To specify a checkbox field use: `type: "checkbox"`
+The `prefill` value can determine the default state of the checkbox and it is required.
+
+```js
+var options = {
+  additionalSignUpFields: [{
+    type: "checkbox",
+    name: "newsletter",
+    prefill: "true",
```
selaux: Is there any reason for this being a string instead of a boolean value?

dariobanfi: I wanted to keep the modifications as small as possible.
The prefill value is needed otherwise the checkbox will be "invalid" in the store and the registration won't go through. 
Prefill must be string: https://github.com/auth0/lock/blob/master/src/connection/database/index.js#L107
I could change that, but it would need a lot of other changes elsewhere in the code

PR #876, opened 2017-02-21
Updating jest and snapshots

luisrudge:


luisrudge:
I'll wait for the other PR's before updating the snapshots so we don't end up in a rebase hell

PR #889, opened 2017-03-01
Updated auth0.js to v8.3.0

### LINKED ISSUE #867: Doesn't work in cordova
I use `auth0-lock` v10.11.0, in browser it's work well, but when i tried use in cordova, there is error, after inAppBrowser open and request permision, it's close, but auth0 pop still spinning, and when i debug, appear this :

screenshot from 2017 02 09 12 10 00

maybe anyone can suggest solution to me,

Thx

hzalaz:
Start using tilde (~) for auth0.js version to get patches automatically.

Fixes #867 by including https://github.com/auth0/auth0.js/pull/370

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "zuul-ngrok": "4.0.0"
   },
   "dependencies": {
-    "auth0-js": "8.2.0",
+    "auth0-js": "~8.3.0",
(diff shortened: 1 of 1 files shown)


ihsansatriawan:
when this version will be available in npm ? because, i still get `auth0-js` version 8.2.0

PR #915, opened 2017-03-10
do not show popup if sso false

### LINKED ISSUE #912: Stack overflow error in lock.js when signing up with user/password
- [ ] Code snippet or sample project that reproduces the bug
Not sure which code would be helpful, but I can potentially provide some.

- [x] Screenshots when appropriate
screen shot 2017-03-08 at 2 45 36 pm

- [x] Lock version
10.12.3 (though I tried many versions 10.x+)

- [x] Browser & OS
Chrome 56.0.2924.87 (64-bit)
Firefox 52.0 (64-bit)
Safari 10.0.3 (12602.4.8)
MacOS 10.12.3 (16D32)

More info:
---
Some background, we were just updating from Lock 9.x to 10.x. Everything mostly working except signup with user/password. The error being provided, if related to our configuration, is very unclear. I tried a bunch of versions of 10.x, even went back as far as 10.4, but it appears going that far back made other things not work.

The error from Chrome dev console is in the screenshot above. Firefox doesn't handle it as gracefully, and mostly hangs until it shows a message asking you if you want to close the tab. Safari just hangs. 

The error appears to be related to a `toSnakeCase` function recursing too much. I can't tell based on the error if we configured something incorrectly, but the rest of things seem to work just fine.

Either way, if we did configure this incorrectly, failing this way is probably a bug.

maxbeatty:
Trying to fix #912 

From the [popup docs](https://auth0.com/docs/libraries/lock/v10/popup-mode#database-connections-and-popup-mode), my understanding is that setting `sso: false` should prevent a popup. Right now, the popup is dictated by `redirect: false`.

Ultimately, I'd like to sign up a user without redirecting and without a popup.

### DIFF
--- a/src/core/web_api/legacy_api.js
+++ b/src/core/web_api/legacy_api.js
@@ -76,7 +76,7 @@ class Auth0LegacyAPIClient {
 
     delete options.autoLogin;
 
-    const popupHandler = (autoLogin && popup) ? this.client.popup.preload() : null;
+    const popupHandler = (autoLogin && sso) ? this.client.popup.preload() : null;
 
     this.client.signup(options, (err, result) => cb(err, result, popupHandler));
   }


hzalaz:
Thanks for the PR but but this will break our API since what tells lock if its popup or not is `redirect` option so we won't merge this. 

The fix to the underlying issue will be done in auth0.js soon

maxbeatty:
Can you point me to the fix being made in auth0.js? 

I'm a paying customer who can't sign up new users. I'd like to do whatever I can to resolve this as quick as possible.

hzalaz:
@maxbeatty I can't since as I said before will be done so we haven't started on it but it should be available next week
You could probably downgrade Lock to 10.8.x as a quick workaround until we release a new lock version since the toSnakeCase is part of our update of auth0.js and was added in 10.9.x.

Also as a paying customer I'd strongly recommend using https://support.auth0.com since we don't check 24/7 github issues and we can't ensure an SLA, provide quick support or even tell paying customers from non-paying ones.

maxbeatty:
I submitted a support ticket shortly after opening the PR. You've been faster to respond here on GitHub and it's much appreciated.

PR #924, opened 2017-03-13
Updated auth0 js version

### LINKED ISSUE #896: auth0-lock 10.12.1 breaking
Version 10.12.1 of auth0-lock creates an error:
```
Uncaught Error: clientID is required.
    at required (assert_required.js:17)
    at new Auth0 (index.js:142)
    at ~~our_client_id_here~~:526
    at ~~our_client_id_here~~:416
```
 in the auth0 pop-up after logging in from our site.

The error (which is wicked hard to catch since it happens on the pop up) happens near line 399 of the generated javascript at the auth0.com url that resembles `https://ourcompany.auth0.com/sso_dbconnection_popup/~~our_client_id_here~~`:

```javascript
function onMessage(e) {
          // only one message gets through, but let's make sure it's actually
          // the message we're looking for (other code may be using
          // postmessage) - we do this by ensuring the payload can
          // be parsed, and it's got an 'a' (action) value of 'request'.
          var d;
          try {
            d = JSON.parse(e.data);
          } catch(err) { }
          if (!d || d.a !== 'request') return;
          removeListener(window, 'message', onMessage);
          o = e.origin;
          if (cb) {
            // this setTimeout is critically important for IE8 -
            // in ie8 sometimes addListener for 'message' can synchronously
            // cause your callback to be invoked.  awesome.
            setTimeout(function() {
              cb(o, d.d, function(r) {
                cb = undefined;
                doPost({a: 'response', d: r});
              });
            }, 0);
          }
        }
```

The value `d.d` is undefined, which is usually the object that carries the clientId value. Only `d.a` gets set to anything (`'request'`). Otherwise, the `d` object is empty.

It's a little difficult to describe this error any more beyond this, but reverting back to 10.11.0 solved the issue for us. I saw issues on here from a year or so ago that showed similar errors due to something with minification, and I noticed that 10.12.1 included transpilation changes, so that _may_ be the culprit here.

Hope that helps!

### LINKED ISSUE #912: Stack overflow error in lock.js when signing up with user/password

hzalaz:
This should include fixes for popup issues when using db connections.

Closes #896 and fixes #912


Floby:
Is this fix already available in a published version ?

PR #945, opened 2017-03-24
Don't clear email field after reset password

### LINKED ISSUE #933: prefill option is lost after reset password
The prefill option does not carry through the reset password flow, so after requesting a password reset, Auth0 redirects back to the login screen but now with an empty email field. I would expect email to still be filled out, to minimize the chance of typos in the email field during login (which is why we are using prefill in the first place). See the screen shots below for details.

The issue can be reproduced with the HTML code below. Note that the clientId / domain combination that I use is configured for "Username-Password-Authentication" connection.

Browser: Seen in Chrome
OS: Seen on Windows

```
<script src="https://cdn.auth0.com/js/lock/10.11.0/lock.min.js"></script>
<script>
var clientId = '(my clientId)';
var domain = '(my domain)';
var options = {
  allowedConnections: [
    'Username-Password-Authentication'
  ],
  prefill: {
    email: 'lars@example.com'
  }
};
var lock = new Auth0Lock(clientId, domain, options);
lock.show();
</script>
```
login-prefill-before-reset
login-prefill-reset
login-prefill-after-reset

luisrudge:
fix #933

### DIFF
--- a/src/connection/database/actions.js
+++ b/src/connection/database/actions.js
@@ -181,7 +181,7 @@ function resetPasswordSuccess(id) {
   const m = read(getEntity, "lock", id);
   if (hasScreen(m, "login")) {
     swap(updateEntity, "lock", id, m => (
-      setScreen(l.setSubmitting(m, false), "login")
+      setScreen(l.setSubmitting(m, false), "login", [""])
     ));
 
     // TODO: should be handled by box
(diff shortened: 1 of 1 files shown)


src/connection/database/actions.js
```diff
@@ -181,7 +181,7 @@ function resetPasswordSuccess(id) {
   const m = read(getEntity, "lock", id);
   if (hasScreen(m, "login")) {
     swap(updateEntity, "lock", id, m => (
-      setScreen(l.setSubmitting(m, false), "login")
+      setScreen(l.setSubmitting(m, false), "login", [""])
```
hzalaz: Add a comment why there is a `[""]` as a parameter since we will quickly forget and maybe add a `FIXME` so we fit it eventually

PR #946, opened 2017-03-24
Clarify `sso` option behavior in the README

selaux:
As stated in [this support ticket](https://support.auth0.com/tickets/20698) the sso switch does not actually influence the creation of a SSO session on the Auth0 side. This is imho not properly reflected in the docs for this switch, so I updated it.

I'm still not sure if the `sso` switch in the current state couldn't be unified with `rememberLastLogin` because, according to the ticket, both are only frontend features and the SSO data is probably not used for other cases.


luisrudge:
@selaux I improved the description of the flag a bit and opened another PR: https://github.com/auth0/lock/pull/948

PR #949, opened 2017-03-24
disable social buttons when terms were not accepted on sign up

### LINKED ISSUE #919: Block/disable social buttons when mustAcceptTerms is not selected
We need to prevent a user from clicking the social buttons or starting the operation if the terms checkbox is not selected. 
Some alternatives are:

- fail with error `must accept terms`
- disable the buttons

Also we need to update docs stating is not only for DB connections

luisrudge:
fix #919 

disable-social

### DIFF
--- a/css/index.styl
+++ b/css/index.styl
@@ -440,6 +440,8 @@ tabsHeight = 40px
     position relative
     width 40px
     height 40px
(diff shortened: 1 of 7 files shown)


luisrudge:
cc @vctrfrnndz

PR #962, opened 2017-03-31
Fix background blur in the header

### LINKED ISSUE #954: Lock header inconsistent across browsers
I am using latest version of Auth0 Lock (v10). When specifying custom theme color like so:

```
options = { 
         ...
        theme: {
            primaryColor: '#3A99D8',
            labeledSubmitButton: false
        },
         ...
     }
```

The Lock header background shows up with blurred version of primaryColor in all browsers except Chrome.

I noticed that div with class="auth0-lock-header-bg-solid" is not positioned correctly on the Lock header in Chrome while in other browsers it gets placed correctly. In Chrome, an extra css class "auth0-lock-blur-support" gets added to the div above the div.auth0-lock-header-bg-solid which pushes it down.

It seems the Lock on auth0.com's homepage also has same behavior. 

Is there a way we can turn off this blur-support without overriding css locally?

Please see the screenshots below:

other-browsers

chrome

Thanks.

Riyaz

luisrudge:
fix #954

### DIFF
--- a/css/index.styl
+++ b/css/index.styl
@@ -238,11 +238,6 @@ tabsHeight = 40px
         background-position center
         background-size 800px
 
-    &.auth0-lock-blur-support
-      .auth0-lock-header-bg-blur
-        +breakpoint("mobile")
-          display block
-
     .auth0-lock-header-bg-solid
       height 100%
       opacity .08


beneliflo:
This issue #954 isn't a issue. All lock have a blurred header of his icon/logo in grayscale. The problem is IE who don't support blur property and the fix for that is just add the color primary with opacity instead. Mobile works like IE so we can't delete this css line because we will be broken.
I think the solution is make a option for this and let the client choose.

PR #967, opened 2017-04-04
Use replaceState for better browser history experience

selaux:
__Rationale__: Currently, when we get to the login callback, the callback url with and without the hash will be counted as two different browser history entries. This leads to the user not having the possibility to navigate back as the history entry that includes the hash will always trigger the authentication procedure again. Additionally I don't like the idea of tokens appearing in my browser history, because this way an attacker could steal them from there (although they will only be valid for a limited amount of time).

__Implementation__: Replace `window.location.hash = ''` with `window.history.replaceState(` so it will not create an additional history entry, but replace the existing one that includes the hash.

What do you think?

### DIFF
--- a/src/core/actions.js
+++ b/src/core/actions.js
@@ -29,10 +29,11 @@ export function setupLock(id, clientID, domain, options, hookRunner, emitEventFn
 export function handleAuthCallback() {
   const ms = read(getCollection, "lock");
   const keepHash = ms.filter(m => !l.hashCleanup(m)).size > 0;
+  const urlWithoutHash = global.location.href.split('#')[0];
   const callback = (error, authResult) => {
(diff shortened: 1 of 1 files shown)


luisrudge:
I don't see two history entries when I log in on Chrome. 🤔

selaux:
Tested on Firefox and Chrome, but maybe I didn't describe it correctly and maybe this is a special case that we have. I will try to build a minimal example to showcase the issue.

selaux:
Cool, thank you!

selaux:
Thinking back on it you might also need to consider browser support. E.g. IE9 does not support this.

luisrudge:
I saw that, but we don't support IE9, so that's ok

PR #971, opened 2017-04-14
Upgrading usage of prop-types to new package

### LINKED ISSUE #970: Proptypes in the main react package is deprecated
`Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.`

see https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html

Please start using prop-types in the next update so the error goes away, thanks

luisrudge:
fix #970


luisrudge:
Yes! https://github.com/reactjs/prop-types#compatibility
**React 0.14.9** and  **React 15.3.0** and higher

PR #973, opened 2017-04-15
fixed propType misspell in header.jsx

nickpisacane:
Change propType `backgorundColor` to `backgroundColor`. Little bit of shameless self-promotion here, but I caught this using [maybe-you-meant](https://github.com/nickpisacane/maybe-you-meant) while working on an app of mine. Appreciate the great work on Lock!


luisrudge:
Thanks for the PR! 🎉

PR #997, opened 2017-05-09
Fixed overrides sent to auth0.js

sandrinodimattia:
Overrides are currently not working, because auth0.js requires them to be part of `opts.overrides`, while currently they are being added on `ops` directly.

Reference: https://github.com/auth0/auth0.js/blob/master/src/web-auth/index.js#L48-L53


hzalaz:
Needs a release of auth0.js to be merged

Issue #878, opened 2017-02-23
Error messages do not allow html formatting


dariobanfi:
Hi,

Is there any reason why error messages displayed by the lock do not allow html formatting?

For example, given this lock parameters:

```
error: {
  login: {
     'lock.invalid_email_password': '<b>Error</b>, Invalid Email or Password'
  }
}
```

Which results as:

![screen shot 2017-02-23 at 17 06 25](https://cloud.githubusercontent.com/assets/4538567/23267242/7102e976-f9ea-11e6-8365-281f7ed3ae04.png)

This could be easily changed by using `i18n.html` here: https://github.com/auth0/lock/blob/master/src/core/index.js#L511 and it's what most of other messages do (for example `success.signUp`).

luisrudge:
Thanks @dariobanfi. I added a PR to support html messages (both error and success). We'll discuss internally if we want to support more stuff as html or not, but this should be in the next release.

PR #1017, opened 2017-05-29
Enabled HTML formatting for flashMessages

dariobanfi:
As part of https://github.com/auth0/lock/issues/878#issuecomment-286882027, successfully merged through https://github.com/auth0/lock/pull/928 you enabled html formatting in error/success messages.

However this only work only through languageDictionary strings and not when the message is called through: `lock.show({flashMessage: {...})`
 
This small change should fix that.

### DIFF
--- a/src/__tests__/ui/box/__snapshots__/global_message.test.jsx.snap
+++ b/src/__tests__/ui/box/__snapshots__/global_message.test.jsx.snap
@@ -1,0 +1,31 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`GlobalMessage renders correctly 1`] = `
+<div
+  className="auth0-global-message auth0-global-message-success"
+>
+  <span
+    className="animated fadeInUp"
+    dangerouslySetInnerHTML={
+      Object {
+        "__html": "Success!",
+      }
+    }
+  />
+</div>
+`;
+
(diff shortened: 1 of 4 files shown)


dariobanfi:
(Failing tests shouldn't be because of my change I think... I just cloned the latest master)

luisrudge:
I triggered a rebuild. In the meantime, can you add a test for this? Thanks for the PR!

dariobanfi:
@luisrudge I added some component tests

luisrudge:
Hi @dariobanfi the tests are failing because you changed the way you render errors.
We validated [https://github.com/auth0/lock/blob/f82d025fd9015d7e381174657f38e850d6dd1e69/test/mfa_ro.ui.test.js#L85](here) that the error was shown using [this](https://github.com/auth0/lock/blob/ba69824e09329dec95a50bd98172eb2e4d0f7350/test/helper/ui.js#L195) helper function. You should make sure the tests reflect the change you made and they pass. Let me know if you need more help. Thanks again for the PR!

dariobanfi:
I updated the PR.
The problem with tests was caused by an inconsistency on how GlobalMessage `message` prop was passed, sometimes it was a React component (when using i18n strings) and sometimes a string (when using flashMessage).
I changed so that it's always passed as React component.

luisrudge:
Thanks for the PR @dariobanfi 🎉

Issue #959, opened 2017-03-31
Scroll to error message when not in viewport


selaux:
We noticed that on smartphones it happens that when you try to login an have invalid credentials, the error message might appear outside the very small viewport that you have when the on-screen keyboard is shown. So the user basically gets no feedback on his login attempt. This could be fixed by scrolling to the lock container top when an error message is shown outside the viewport. What do you think?

luisrudge:
maybe we can instead remove the focus from the keyboard and (hopefully) close it?

luisrudge:
fixed by #1023

PR #1023, opened 2017-06-06
Scroll messages in all lock dialogs into view

m-idler:
Messages might appear outside of users viewport. This PR fixes this by scrolling global messages into the users viewport. Hiding the users keyboard doesn't fix the problem because of the huge amount of different resolutions and display ratios.
see https://github.com/auth0/lock/issues/959

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -156,7 +156,7 @@ function extractUIOptions(id, options) {
     allowAutocomplete: !!options.allowAutocomplete,
     authButtonsTheme: typeof authButtons === 'object' ? authButtons : {},
     scrollGlobalMessagesIntoView: undefined === options.scrollGlobalMessagesIntoView
-      ? false
+      ? true
       : !!options.scrollGlobalMessagesIntoView
   });
 }
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -154,7 +154,10 @@ function extractUIOptions(id, options) {
     primaryColor: typeof primaryColor === 'string' ? primaryColor : undefined,
     rememberLastLogin: undefined === options.rememberLastLogin ? true : !!options.rememberLastLogin,
     allowAutocomplete: !!options.allowAutocomplete,
-    authButtonsTheme: typeof authButtons === 'object' ? authButtons : {}
+    authButtonsTheme: typeof authButtons === 'object' ? authButtons : {},
+    scrollGlobalMessagesIntoView: undefined === options.scrollGlobalMessagesIntoView
+      ? false
+      : !!options.scrollGlobalMessagesIntoView
   });
 }
 
@@ -185,7 +188,8 @@ export const ui = {
   primaryColor: lock => getUIAttribute(lock, 'primaryColor'),
   authButtonsTheme: lock => getUIAttribute(lock, 'authButtonsTheme'),
   rememberLastLogin: m => tget(m, 'rememberLastLogin', getUIAttribute(m, 'rememberLastLogin')),
-  allowAutocomplete: m => tget(m, 'allowAutocomplete', getUIAttribute(m, 'allowAutocomplete'))
+  allowAutocomplete: m => tget(m, 'allowAutocomplete', getUIAttribute(m, 'allowAutocomplete')),
+  scrollGlobalMessagesIntoView: lock => getUIAttribute(lock, 'scrollGlobalMessagesIntoView')
 };
 
 const { get: getAuthAttribute } = dataFns(['core', 'auth']);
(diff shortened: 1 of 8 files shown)


luisrudge:
Can you make a gif comparing both results?

m-idler:
Current state - you won't recognise any errors that occur
state-current
changed state - lock dialog will scroll to the position of the error message
state-new

In the previous commit we added a parameter ```scrollGlobalMessagesIntoView``` to not brake/change current behaviour. By default  ```scrollGlobalMessagesIntoView``` is disabled.

src/core/index.js
```diff
@@ -183,7 +186,8 @@ export const ui = {
   popupOptions: lock => getUIAttribute(lock, 'popupOptions'),
   primaryColor: lock => getUIAttribute(lock, 'primaryColor'),
   authButtonsTheme: lock => getUIAttribute(lock, 'authButtonsTheme'),
-  rememberLastLogin: m => tget(m, 'rememberLastLogin', getUIAttribute(m, 'rememberLastLogin'))
+  rememberLastLogin: m => tget(m, 'rememberLastLogin', getUIAttribute(m, 'rememberLastLogin')),
+  scrollGlobalMessagesIntoView: lock => getUIAttribute(lock, 'scrollGlobalMessagesIntoView')
```
luisrudge: shouldn't this be `tget(m, 'scrollGlobalMessagesIntoView', getUIAttribute(m, 'scrollGlobalMessagesIntoView')),`?

src/core/index.js
```diff
@@ -211,7 +215,8 @@ function extractAuthOptions(options) {
     sso,
     state,
     nonce
-  } = options.auth || {};
+  } =
```
luisrudge: no need to break line here

luisrudge:
That's amazing. Thanks for the PR 🎉

src/core/index.js
fstoerkle: what's the difference? actually, I am not really sure what `tget`,`tgetUI`, etc are doing. However, all properties above don't use `tget`, so I would assume the current way to retrieve the option is sufficient.

luisrudge: Yup. you're right. I was just looking at the rememberLastLogin option, but it is a special case.

luisrudge:
Alright, we're good. I just requested some feedback from our ui/ux designer. I'll wait for him before merging this. Thaaaaaaanks for the PR 👍

src/core/index.js
fstoerkle: 👍

src/core/index.js
m-idler: Seems like this was changed by prettier

m-idler:
Add to small additional commits to fix one of the new tests and add some documentation for the new parameter to the README

luisrudge:
Hey, can you rebase this? THanks!

beneliflo:
I just tested and what I saw is the scroll it varies depends of the height. This happen too with you?

lock_scroll

luisrudge:
I think you have to refresh after resizing the height, no?

beneliflo:
@luisrudge with refresh. Same:

lock_scroll_2

luisrudge:
thanks @beneliflo. Can you take a look @m-idler?

m-idler:
I can reproduce the mentioned behaviour with small viewport heights too. After playing around with it a little bit, it seems that it's not caused by the scrolling itself but by the focus being set to the first input later.
In chrome.jsx I changed the timeout value from 17 to 1000 for testing:
```javascript
if (!prevProps.error && error) {
      const input = this.findAutofocusInput();
      if (input) {
        // TODO clear timeout
        setTimeout(() => input.focus(), 1000);
      }
      return;
}
```
For chrome, it seems that the focused input is vertically-center-scrolled into the viewport, in firefox it's just scrolled at any position into the viewport. You can see the behaviour of Chrome and Firefox in the attached gifs.

Im not sure if it's a (good) solution, to skip setting the focus when ```scrollGlobalMessagesIntoView``` is enabled

chrome-focus
firefox-focus

luisrudge:
Thanks @m-idler. Just to give you a feedback, I'm discussing internally with our design team if we should keep the scroll-to-input behaviour at all. I'm also discussing if we should make `scrollGlobalMessagesIntoView` default to `true` instead of `false`. I'll push to your branch once I receive the final answer.

luisrudge:
Thanks @m-idler for all your effort. I made a small change to make this the default behaviour and removing the scroll-to-input behaviour-on-error that you pointed out. Starting from the next release, `scrollGlobalMessagesIntoView` will be true by default. 🎉

luisrudge:
🎉 :shipit: 🚀

PR #1051, opened 2017-06-29
Created korean language file for supporting korean users

couldseeme:
Hello. I'm a freelancer developer based in Seoul, Korea.
In this PR, I added korean language file which is for supporting korean users.
I thought it would be great if there is a korean version of Auth0-lock, so I send you this PR.
I hope you like it. Thank you.


luisrudge:
Thanks @couldseeme 🎉

PR #1052, opened 2017-06-29
Adding a custom connection resolver option

### LINKED ISSUE #937: Feature Request: Extensible Connection Name Resolution (HRD)
**Scenario:**

* You already have an Auth0 solution that leverages the Lock widget to facilitate logon
* You utilize the Home Realm Discovery feature to build a slick auth experience for users who have their own Enterprise IdP like Azure AD
* For those users that don't bring their own IdP you want to be able to place them in separate database connections 
  * Perhaps by e-mail domain or some other arbitrary rule

**Problem**
Unfortunately today the Lock Widget does not support a Home Realm Discovery experience for Database users. I can see why that might be.

Database connections support a number of different permutations that don't all lend themselves naturally to HRD and so you'd probably need to invest a reasonable amount of time and resources into making that work.

But many of your customers (myself included) would probably not have such open requirements.
We might be able to determine connection name from:

- A naming convention (by e-mail domain for example)
- An API request
- etc

**Proposal**
From a layman's perspective HRD in Auth0 could be summarized as 
> "Given a username and some context resolve the name of the connection you'd like to authenticate against"

Could you build an extensibility point into the Lock widget that allows us to inject our own logic here?
Something similar to the below...
```js
// This is hand written and probably contains all manner of syntax errors - be nice :P
var lock = new Auth0Lock(client, domain, options)
lock.show({
      icon: 'foo.png',
      callbackURL: '/callback',
      responseType: 'code',
      connectionResolver: function(username, context) {
          var domain = username.split("@")[1];
          var chosenConnection = context.connections.find(c => c.domains.indexOf(domain) >= 0);
          // or perhaps...
          var chosenConnection = context.connections.find(c => c.name == domain);
          // Promises to support alternative async scenarios
          return Promise.resolve(chosenConnection.name);
      }
    });
```
Is this an area you would be interested in exploring further?

luisrudge:
fix #937

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,6 +292,23 @@ var options = {
 - **configurationBaseUrl {String}**: Overrides client settings base url. By default it uses Auth0's CDN url when `domain` has the format `*.auth0.com`. Otherwise, it uses the provided `domain`.
 - **languageBaseUrl {String}**: Overrides the language source url for Auth0's provided translations. By default it uses to Auth0's CDN url `https://cdn.auth0.com`.
 - **hashCleanup {Boolean}**: When enabled, it will remove the hash part of the callback url after the user authentication. Defaults to `true`.
+- **connectionResolver {Function}**: When in use, provides an extensibility point to make it possible to choose which connection to use based on the username information. Has `username`, `context` and `callback` as parameters. The callback expects an object like: `{type: 'database', name: 'connection name'}`. **This only works for database connections.** Keep in mind that this resolver will run in the username/email input's `onBlur` event, so keep it simple and fast.
+
+```js
+var options = {
+  connectionResolver: function (username, context, cb) {
+    var domain = username.includes('@') && username.split('@')[1];
+    if (domain) {
+      // If the username is test@auth0.com, the connection used will be the `auth0.com` connection.
+      // Make sure you have a database connection with the name `auth0.com`.
+      cb({ type: 'database', name: domain });
+    } else {
+      // Use the default approach to figure it out the connection
+      cb(null);
+    }
+  }
+}
+```
 
 #### Language Dictionary Specification
 
--- a/src/__tests__/field/__snapshots__/email_pane.test.jsx.snap
+++ b/src/__tests__/field/__snapshots__/email_pane.test.jsx.snap
@@ -16,6 +16,7 @@ exports[`EmailPane renders correctly 1`] = `
   data-autoComplete={false}
   data-invalidHint="invalidErrorHint"
   data-isValid={false}
+  data-onBlur={[Function]}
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -292,6 +292,23 @@ var options = {
 - **configurationBaseUrl {String}**: Overrides client settings base url. By default it uses Auth0's CDN url when `domain` has the format `*.auth0.com`. Otherwise, it uses the provided `domain`.
 - **languageBaseUrl {String}**: Overrides the language source url for Auth0's provided translations. By default it uses to Auth0's CDN url `https://cdn.auth0.com`.
 - **hashCleanup {Boolean}**: When enabled, it will remove the hash part of the callback url after the user authentication. Defaults to `true`.
+- **connectionResolver {Function}**: When in use, provides an extensibility point to make it possible to chose which connection to use based on the username information. Has `username`, `context` and `callback` as parameters. The callback expects an object like: `{type: 'database', name: 'connection name'}`. **This only works for database connections.**
```
hzalaz: `chose` -> `choose`

README.md
```diff
@@ -292,6 +292,23 @@ var options = {
...
+var options = {
+  connectionResolver: function (username, context, cb) {
+    var domain = username.includes('@') && username.split('@')[1];
+    if (domain) {
+      // If the username is test@auth0.com, the connection used will be the `auth0.com` connection.
+      // Make sure you have a database connection with the name `auth0.com`.
+      cb({ type: 'database', name: domain });
+    } else {
+      // Use the default approach to figure it out the connection
+      cb(undefined);
```
hzalaz: I'd rather use `cb(null)` or even `cb()` and we should handle both cases gracefully

hzalaz: Also in the implementation lets add a warning or timeout if this operation takes too long (could be in a diff PR as long as we add a note saying Keep It Simple)

README.md
hzalaz: The name `connectionResolver` is ok as long we don't have a provider or similar concept elsewhere. If we do lets be consistent 😄

luisrudge: only gravatar provider, but it's not part of the public api.

README.md
luisrudge: it doesn't really matter what you set here.. null, empty, undefined, false.. everything that evaluates to `false` will make it behave the same.
I'll add a note in the readme because the timeout will come back to bite us for sure

hzalaz: Ok but let's remove the `cb(undefined)` from he example, and that we have a good error when the object you pass is empty or has a wrong format

README.md
hzalaz: 👍

PR #1073, opened 2017-08-07
Add Slovenian translations

jogee:


luisrudge:
Thanks again! 🎉

urosgruber:
If this is Slovenian language then I must be from other planet. Can you please revert this as it doesn't make any sense.

urosgruber:
Actually Romanian and Slovenia files are named in wrong way.

luisrudge:
@urosgruber did you check this PR? https://github.com/auth0/lock/pull/1092

urosgruber:
@luisrudge no, but looks like all good now.

PR #1076, opened 2017-08-08
Fix spacing using custom signup fields

### LINKED ISSUE #1075: Lack of spacing in Lock 10.19 additional signup fields
When adding additional signup fields to Lock 10.19 or Lock 10.18, there is a lack of space between a couple of fields, which looks odd. If lock 10.17 is used then it works fine

You can see a screenshot of the issue here:
image

Lock 10.17 version here:

image

luisrudge:
fix #1075


luisrudge:
before:
image


after:
image

Issue #1090, opened 2017-08-20
Problematic comparison in store : observe() : subscribe()


UnsungHero97:
Auth0 Lock v10.20.0

I'm using an undocumented API method, `lock.destroy()`, and I encountered issue https://github.com/auth0/lock/issues/907. After tracing the code, I think I've found a problematic comparison that leads to that issue:

https://github.com/auth0/lock/blob/7f0e0b59a9785d2b887d24591ebea2e61f2d69d8/src/store/index.js#L8-L10

Once `destroy()` is invoked, `getEntity()` returns `undefined` for both `m` and `oldM`, which means `f(m)` will be invoked with `m` being `undefined`, which then causes `Uncaught TypeError: Cannot read property 'getIn' of undefined` in https://github.com/auth0/lock/issues/907.

Shouldn't this check instead be:

`if (m && oldM && m != oldM) f(m);`

...? Furthermore, since both `m` and `oldM` are both `immutable-js` Maps, shouldn't `m != oldM` instead be `!m.equals(oldM)`?

UnsungHero97:
@luisrudge, here you go!

https://github.com/auth0/lock/pull/1091

luisrudge:
fixed in #1653

Issue #907, opened 2017-03-07
Unable to destroy Lock instance


sandrinodimattia:
Hey guys,

I see this error when I call `lock.destroy()`:

```js
Uncaught TypeError: Cannot read property 'getIn' of undefined
    at http://localhost:3001/assets/vendors.js:16215:15
    at getUIAttribute (http://localhost:3001/assets/vendors.js:7967:10)
    at Object.avatar (http://localhost:3001/assets/vendors.js:7984:12)
    at http://localhost:3001/assets/vendors.js:6961:25
    at Object.render-1 (http://localhost:3001/assets/vendors.js:12553:20)
    at http://localhost:3001/assets/vendors.js:15998:24
    at Array.forEach (native)
    at Atom._change (http://localhost:3001/assets/vendors.js:15997:26)
    at Atom.swap (http://localhost:3001/assets/vendors.js:15969:17)
    at swap (http://localhost:3001/assets/vendors.js:12566:21)
    at removeLock (http://localhost:3001/assets/vendors.js:7289:19)
    at Auth0Lock.destroy (http://localhost:3001/assets/vendors.js:7061:29)
    at Auth0Lock.<anonymous> (http://localhost:3001/assets/client.js:35:12)
    at Auth0Lock../node_modules/events/events.js.EventEmitter.emit (http://localhost:3001/assets/vendors.js:31155:17)
    at http://localhost:3001/assets/vendors.js:8337:35
```

luisrudge:
@sandrinodimattia can you share more about this? what lock version, how are you instantiating lock, how are you calling lock.show()?

luisrudge:
I can repro this. We'll fix this eventually. Since it's not a documented API, I'm assigning this with a low priority

iorrah:
Hey @luisrudg, just landed on this thread from March 2017 and was wondering if this will be tackled any time soon?

luisrudge:
When the component is unmounted, GC will get rid of your internal state, so there's no memory leaks. Yes, we still have to fix this and there's a PR to fix it: https://github.com/auth0/lock/pull/1091 - but this PR touches in a crucial part of the widget and I didn't have time to test all features of lock to make sure everything works.

jkhlim:
Just want to add a +1 to this feature.

luisrudge:
dupe #1090

PR #1091, opened 2017-08-21
fix "Uncaught TypeError: Cannot read property 'getIn' of undefined"

UnsungHero97:
fix for https://github.com/auth0/lock/issues/1090, https://github.com/auth0/lock/issues/907


luisrudge:
Hi @UnsungHero97 can you take a look on those failing tests?

UnsungHero97:
@luisrudge, I looked at it, made a minor change, and the tests are now passing.

luisrudge:
Thanks! I'm not super into immutable, can you please explain what is going on and what you did to fix it?

UnsungHero97:
Sure.

The old comparison was `m != oldM`, which makes sense conceptually ("are these values different?"), but it doesn't make sense in practice because `m` and `oldM` are objects, and more specifically, instances of `Immutable.Map`. If my interpretation of this bit of code is correct, then this is meant to check equality between two objects, and `!=` is not enough to make that decision.

Thankfully, `immutable-js`offers `.equals()`:

https://facebook.github.io/immutable-js/docs/#/Map/equals
https://facebook.github.io/immutable-js/docs/#/is

luisrudge:
Great, thanks. Do you think we need to add another test for this case?

UnsungHero97:
I don't know your codebase well enough, but from the little digging I did, it looks like `m` is guaranteed to be an instance of an `immutable-js` object, so `.equals()` should be available.

If there's test coverage for your use of `immutable-js`, then I think  we're safe. If not, it's probably a good idea to add tests to ensure we're working with `immutable-js` instances.

luisrudge:
alright. thanks man! This will probably ship in the next release 🎉

aautem:
Any update on this?

luisrudge:
new pr: https://github.com/auth0/lock/pull/1091

Thanks @UnsungHero97 for fixing this and I'm super sorry we couldn't merge this before.

PR #1092, opened 2017-08-22
Fixed misplaced translations for Romanian and Slovenian

AdrianSima:
- swapped Romanian and Slovenian translations
- fixed a few strings for Romanian translation


luisrudge:
Can you explain what happened?

AdrianSima:
I got here because I get an error when setting "language = 'ro'" on the Hosted Pages for the Lock widget.
After looking in the file 'src/i18n/ro.js' I saw that it was actually the wrong language. After using Goggle Translate, found out that it was the Slovenian one.

Then I went to 'src/i18n/sl.js' and found the Romanian translation there. I did a fork to swap those contents and while at it fixed a few strings for the Romanian language.

luisrudge:
Thanks! That make sense. Not sure why languages were switched.

jogee:
Must've mixed them up in the process, not a native speaker, sorry for the confusion.

brgrz:
However, the Slovenian translations are missing Slovenian special charasters (like š,č,ž). I though it was a html charset issue (I always set it to uft-8) but apparently the characters are actually missing in the translation file.

brgrz:
Made another fix to the Slovenian translations, please approve and merge

https://github.com/auth0/lock/pull/1171

PR #1097, opened 2017-09-03
Improve Danish translation

havgry:
I just went through the Danish translation and found two minor mistakes. Overall the translation is pretty accurate.


luisrudge:
Thanks for the PR! 🎉

PR #1100, opened 2017-09-07
Email and user input lack a width and height

KevinBusse:
Email and user input lack a width and height, thus they are easily manipulated by third party CSS like:

```css
svg {
  display: block;
  height: 100%;
  width: 100%;
}
```


luisrudge:
There were a few other places that needed some fixes, so I fixed all of them in a different PR. Thanks for bringing this to our attention! This will be fixed in the next release. 🎉

PR #1087, opened 2017-08-16
Fire connectionResolver onSubmit instead of onBlur

NicoleGrondin:
Fixes #1081

Fires connectionResolver onSubmit instead of onBlur. The connectionResolver function was not getting executed when email and username fields were pre-filled by password managers (LastPass..., Mozilla password manager).

- [x] Removes blur event from username and email pane
- [x] Fires connectionResolver onSubmit
- [x] Updates tests

cc @luisrudge

PR #1111, opened 2017-09-19
Resolve connection for reset password

lukevmorris:
Looks like the `connectionResolver` option was added in 10.19 to allow control over which connection to authenticate against during `logIn`. We've been using this feature since its release, and it's worked well so far.

We've noticed that `resetPassword` does not respect this function, and this can result in some really confusing behavior. For users whose default connection is overridden by the `connectionResolver`, they could be logging into one connection but resetting their password in another. And so their interaction with the Lock might go as follows:

* Log in for the first time in a month and don't really remember their password. They try their default. Incorrect.
* They click "Forgot your password?" and trigger a password reset email. They click on the link in the email and submit a new password.
* They try to log in again with their new password. Incorrect.
* Typically we get a support email after this.

I looked at the original `connectionResolver` PR (#1052) to get my bearings in the project. I was hoping to mirror some test cases for `resetPassword`, but I only saw new pane tests, which I believe should remain unchanged in this PR. Let me know if there are any changes you'd like to see here, and I'll try to update ASAP.

Thanks!

### DIFF
--- a/src/connection/database/index.js
+++ b/src/connection/database/index.js
@@ -288,7 +288,7 @@ export function databaseConnection(m) {
 }
 
 export function databaseConnectionName(m) {
-  return (databaseConnection(m) || new Map()).get('name');
+  return (databaseConnection(m) || Map()).get('name');
 }
 
 export function forgotPasswordLink(m, notFound = '') {
--- a/src/connection/database/index.js
+++ b/src/connection/database/index.js
@@ -282,8 +282,11 @@ export function databaseConnection(m) {
   return defaultDirectory(m) || defaultDatabaseConnection(m) || l.connection(m, 'database');
 }
 
-export function databaseConnectionName(m) {
-  return (databaseConnection(m) || Map()).get('name');
+export function databaseConnectionName(lock) {
+  const customResolvedConnection = l.resolvedConnection(lock);
+  const standardConnection = databaseConnection(lock);
+  const connection = customResolvedConnection || standardConnection || {};
+  return connection.name;
 }
 
 export function forgotPasswordLink(m, notFound = '') {
--- a/src/connection/database/index.js
+++ b/src/connection/database/index.js
@@ -288,8 +288,8 @@ export function databaseConnection(m) {
   );
 }
 
-export function databaseConnectionName(lock) {
-  return (databaseConnection(lock) || new Map()).get('name');
+export function databaseConnectionName(m) {
+  return (databaseConnection(m) || new Map()).get('name');
 }
 
 export function forgotPasswordLink(m, notFound = '') {
--- a/src/connection/database/index.js
+++ b/src/connection/database/index.js
@@ -279,9 +279,8 @@ export function defaultDatabaseConnectionName(m) {
 }
 
 export function databaseConnection(m) {
-  const customResolvedConnection = l.resolvedConnection(lock);
   return (
-    customResolvedConnection ||
+    l.resolvedConnection(m) ||
     defaultDirectory(m) ||
     defaultDatabaseConnection(m) ||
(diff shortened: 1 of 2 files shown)


luisrudge:
Thanks for the PR. Do you think it makes sense to add this to the signup as well?

lukevmorris:
@luisrudge Hmm -- probably. If I think back to when I implemented my `connectionResolver`, I guess I expected it to work throughout the Lock. So it would probably help to clear up some developer confusion in the future if `signUp` used it as well. I'll add that in.

luisrudge:
Great! What did you mean about the tests? Why can't you add tests for this?

lukevmorris:
I'm happy to add tests, I am just not sure where in the project I should add them. Would you be able to direct me?

luisrudge:
actually, now that I'm taking a look at this, [this PR](https://github.com/auth0/lock/pull/1087) will fix the same issue. we're only missing two tests ([here](https://github.com/auth0/lock/pull/1087#discussion_r133526239)) to merge it. Can you help with those tests? I'm focusing on the other PRs right now. 

The PR changes from resolving the connection `onBlur` to `onSubmit` and will make all screens resolve the connection before submitting (will fix forgot password, signup etc). Let's concentrate our efforts in that PR instead, what do you think?

lukevmorris:
I might be wrong, but my current understanding is that this PR is orthogonal to #1087.

It looks like the onBlur / onSubmit callback only sets the resolved connection. But that resolved connection gets ignored unless explicitly asked for in the functions in `database/actions.js`, with `l.resolvedConnection(lock)`. Before this PR, `logIn` calls that function, but `signUp` and `resetPassword` do not.

luisrudge:
You're totally correct. I was taking a look at the PR and thought it was fixing the same issue. Sorry about that.

src/connection/database/index.js
```diff
@@ -282,8 +282,11 @@ export function databaseConnection(m) {
   return defaultDirectory(m) || defaultDatabaseConnection(m) || l.connection(m, 'database');
 }
 
-export function databaseConnectionName(m) {
-  return (databaseConnection(m) || Map()).get('name');
+export function databaseConnectionName(lock) {
+  const customResolvedConnection = l.resolvedConnection(lock);
+  const standardConnection = databaseConnection(lock);
+  const connection = customResolvedConnection || standardConnection || {};
+  return connection.name;
```
luisrudge: You can't directly access `.name` here because it's an Immutable object.
it should be someting like

```js
const connection = customResolvedConnection || standardConnection || new Map();
return connection.get('name');
```

luisrudge: .

lukevmorris:
It looks like I need to push this logic further to the core. There are other functions like `databaseConnectionRequiresUsername` that need access to the resolved connection as well.

src/connection/database/index.js
```diff
@@ -279,11 +279,18 @@ export function defaultDatabaseConnectionName(m) {
...
+    defaultDirectory(m) ||
+    defaultDatabaseConnection(m) ||
+    l.connection(m, 'database')
+  );
 }
 
-export function databaseConnectionName(m) {
-  return (databaseConnection(m) || Map()).get('name');
+export function databaseConnectionName(lock) {
+  const connection = databaseConnection(lock) || new Map();
```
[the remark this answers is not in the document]
lukevmorris: okiedoke

lukevmorris:
Ok -- I believe comments are addressed.

luisrudge:
Thanks for the PR and patience 🎉

lukevmorris:
Thanks very much @luisrudge!

PR #1124, opened 2017-09-27
Make cross origin authentication the default in OIDC mode

luisrudge:


petetnt:
Hi,

Any reason why this wasn't a SemVer major change or does the Lock not follow semver in the first place?

luisrudge:
We follow semver, but `oidcConformant` was never documented. The first documented and supported version is this one with cross origin authentication.

petetnt:
Thanks for the explanation @luisrudge. Was just pretty confused as it's used in (for example) https://github.com/auth0/lock/blob/16cbf992f927b834181e940365f124f2e98fb485/support/index.html#L19 and https://github.com/auth0-samples/auth0-react-samples/blob/embedded-login/01-Embedded-Login/src/Auth/Auth.js#L8 and the supplied samples wouldn't work without it because they contain `auth.audience` option (which we require too).

luisrudge:
Indeed we got a bit ahead of ourselves with the samples. Truth is that it was a huge coordinated effort to release the whole feature and we got a bit out of sync with the samples. In any case,
you wouldn’t find a piece of documentation about it. I understand your frustration and I’m sorry things turned out like this for you. 
We thought we didn’t need to release a new major because it wasn’t actually documented, although there was a few samples around with it. 

Did you find  any particular issues with cross origin authentication? I mean, apart from having to enable stuff in the dashboard?

petetnt:
Only issue we are running into is mainly echoing https://community.auth0.com/answers/9872/view, I don't think there's any way to develop/test under localhost with the current scheme (with some sort of `I understand that this might not be safe nor conformant` clause if necessary)?

The undertaking is understandably huge and we ❤️  the work you do at Auth0! And no worries, my confusion today when things stopped working for a while is nothing compared to trying to do these things from scratch myself.

luisrudge:
Thanks for the kind words ❤️ 

Not sure what you mean about not being able to develop locally. Cross Origin Authentication works just fine locally.

petetnt:
Ach, my bad. It seems that it was due to unrelated issues with Delegated Admin not supporting oicd comformancy. Thanks a ton for help!

luisrudge:
no problem. I'm glad you figured it out.

PR #1132, opened 2017-09-29
Fix defaultUrl sending hash content

luisrudge:


src/core/index.js
```diff
@@ -266,7 +266,7 @@ function extractAuthOptions(options) {
   // if responseType was not set and there is a redirectUrl, it defaults to code. Otherwise token.
   responseType = typeof responseType === 'string' ? responseType : redirectUrl ? 'code' : 'token';
   // now we set the default because we already did the validation
-  redirectUrl = redirectUrl || window.location.href;
+  redirectUrl = redirectUrl || `${window.location.origin}${window.location.pathname}`;
```
lbalmaceda: Will this work on all browsers? Maybe it's better to have a function that **splits** the string in `#` and keeps the first left part.

luisrudge: what part? strings interpolation or using location.origin and pathname?

PR #1139, opened 2017-10-09
remove version 16 from react and react-dom dependencies

whatrocks:
Our app still runs on React 15 but this was installing React 16 as a dependency of the `auth0-lock` package, causing this issue:  `https://reactjs.org/docs/error-decoder.html?invariant=44`. From a user-perspective, this caused issues with logging in, especially when users entered an incorrect password. Removing React 16 here fixed the issue for us.


luisrudge:
Hi, we haven't released lock with react@16 support yet. Are you installing from github?

whatrocks:
Hi @luisrudge - we were using the version from npm and it was still installing react@16 as a dependency of your package.

To fix this, I forked this project, removed the conditional referencing react@16 in your package.json (captured in this PR), and then referenced my fork on GH in our app's package.json. It no longer installed react@16, and the problem went away.

We are using node@6.10.2 if that's helpful.

luisrudge:
can you please paste your package.json file (at least auth0-lock, react and react-dom entries)?
If you have react 15.6.1 ini your package.json file, it should keep it that way.

whatrocks:
`"react-dom": "15.5.4",`
`"react": "15.5.4",`
`"auth0-lock": "10.18.0",`

We are not on react@15.6.1

luisrudge:
so your package.json has a fixed version of react, right? that's pretty weird 🤔

whatrocks:
Yes we have a fixed version in our package.json

luisrudge:
I'll try this out with a brand new create-react-app (pointing to an old version). Are you using yarn or npm?

whatrocks:
Thanks - we're using npm@3.10.10 and node@6.10.2.

luisrudge:
Fixing this on https://github.com/auth0/lock/pull/1142

PR #1142, opened 2017-10-11
Pinning react version 15.6.2

luisrudge:
We can't reliably make this library available in both react@15 and react@16 without changing react and react-dom to peerDependencies. Since this would break everyone moving forward, we decided to pin version ^15.6.2 in this library.


jglover:
Are there plans to release a react 16 compatible package?  
There are drawbacks to using yarn resolutions and for end users the change back and forth is breaking.  
Being locked to react 15 is no good for users who have already (or need to) upgrade. it's putting users in a position to choose between staying on auth0-lock 10.21.x and using yarn resolutions for react.

luisrudge:
What are the drawbacks of using the resolutions field?

jglover:
Probably best covered by the original [RFC](https://github.com/yarnpkg/rfcs/pull/68) & its conversation.  
It's essentially a manipulation on yarn.lock and can have unintended consequences,
 the impression they give is that it's for dire situations like security updates where package authors haven't yet had time to update.  
  
There's no support for it in npm (granted yarn is more widely in use now, and personally I prefer it), but this is forcing end users to use yarn exclusively for now; that's not always desirable/possible. 
It might sound petty but it can leave projects in a position where auth0-lock is the only package preventing them from using conventional dependencies with react 16, if yarn were to deprecate or modify `resolutions` behaviour it could be an issue.  
React 16 has been out almost 6 months, it seems that compatible packages would be a good middle ground for now.

luisrudge:
I understand what you're saying and I can see your point clearly. We understand the trade offs (no npm support or people might not know about yarn resolutions), but we're in a tough spot to solve the issue. We should move to peer dependencies, but we can't do that without releasing a new major version and major reviews of our docs, samples and quickstarts. We just can't dedicate the time to doing that right now. Rest assured we will fix this situation, but, considering we have a clear and simple workaround (resolutions or cdn), we are focusing on more important and valuable things right now. The benefit of waiting a bit more is that more and more projects will upgrade to react@16, which will make our job a bit easier 🎉 . Once we upgrade to react@16, we will suggest people still using react@15 as well, so they can have more time to upgrade.

Again, I'm sorry this is not the answer most of react@16 users want to read, but it is the place that we encounter ourselves right now.

corysimmons:
Happy so long as you guys have it on a roadmap somewhere. 🤤

luisrudge:
We'll absolutely fix this in the next major!

PR #1154, opened 2017-10-20
Adding getProfile method that calls getUserInfo

luisrudge:


src/core.js
```diff
@@ -161,6 +161,9 @@ export default class Base extends EventEmitter {
   getUserInfo(token, cb) {
     return webAPI.getUserInfo(this.id, token, cb);
   }
```
lbalmaceda: the `beautifier` didn't add a new line here. Is that hook working?

luisrudge: it is working, but it doesn't add extra lines. I just added one

PR #1169, opened 2017-11-13
Fixing quick auth with responseType code

luisrudge:
responseType:code requires a redirect. the default way of handling the quick auth is just calling checkSession, which doesn't do a redirect. So, in this case, we need to hit authorize with prompt=none so it performs a redirect.


src/__tests__/core/web_api/p2_api.test.js
```diff
@@ -78,18 +77,6 @@ describe('Auth0APIClient', () => {
         const loginMock = mock.WebAuth.mock.instances[0].login.mock;
         assertCallWithCallback(loginMock, callback);
       });
-      it('should call client.loginWithResourceOwner when redirect is false and sso is false', () => {
```
lbalmaceda: If this premise is no longer valid, negate the test rather than removing it. Helps keep the coverage 👍

PR #1170, opened 2017-11-13
Ensure popupOptions gets passed to auth0.WebAuth

davidrossjones:
As far as I can tell, Lock's `popupOptions` config option is removed by the `normalizeAuthParams` helper and never gets passed to auth0.WebAuth (in either the p2 or legacy apis) where it is used to control the `window.open` settings. Therefore the current `popupOptions` setting listed in the documentation is misleading (it doesn't work)...


luisrudge:
Can you fix the tests? I already updated your branch with latest.

PR #1196, opened 2017-12-15
[RFC] Slightly better Romanian translations

ClaudiuCeia:
Hey there,

The Romanian language uses polite second-person pronouns in formal settings (see https://en.wikipedia.org/wiki/Personal_pronounhttps://en.wikipedia.org/wiki/Personal_pronoun#Formality). The thing is:

- The translations were inconsistent, sometimes the formal version was used, sometimes the informal one
- The audience is different depending on the application folks are building. I'd probably go as far as to say there could probably be two versions of the translation, one for "formal" settings, and one for "informal"? You would't want to address kids/students using a very formal tone, but maybe you would for the customers of a bank, for example? 

**Either way, for now this patch does two things:**
- Fixes the formality inconsistency, the tone used now is informal, not a mix of informal/formal. FWIW, this is the approach most large companies take when it comes to translations (Google, Facebook, etc)
- Fixed some typos here and there as I went through the translated strings. 

**Test plan:**
Honestly I just edited this in the GitHub editor. So uh, eyes. 

Please let me know if there's anything I should do, maybe split this in two variants as suggested above?

### DIFF
--- a/src/i18n/ro.js
+++ b/src/i18n/ro.js
@@ -2,7 +2,7 @@ export default {
   error: {
     forgotPassword: {
       too_many_requests:
-        'Ai depășit limita de introducere a parolei. Avem rugămintea să așteptați până la următoarea încercare.',
+        'Ai depășit limita de introducere a parolei. Avem rugămintea să aștepți până la următoarea încercare.',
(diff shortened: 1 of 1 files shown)


luisrudge:
Thanks for the PR and the clear explanation on what you did! 🎉

PR #1202, opened 2017-12-21
Fix security vulnerability CVE-2017-17068

thoean:
For full details https://auth0.com/docs/security/bulletins/cve-2017-17068.

This upgrades the dependency to the latest version of auth0-js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "zuul-ngrok": "4.0.0"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


luisrudge:
Lock isn't affected by this CVE, because lock itself doesn't use the `popup.callback` method. We'll merge it anyway so we have the latest version though. Thanks 🎉

luisrudge:
released as 10.24.2
Thanks for the PR 🎉

PR #1214, opened 2018-01-03
List Czech as a supported language for auth0/lock.

ivosh:


luisrudge:
Thanks for the PR, but we already have a link for all the available languages.

PR #1224, opened 2018-01-15
`plugins` option added. Forwarded to WebAuth.

lexaurin:
To be able to set plugins during embeded Lock creation. For example:

```
const lock = new Auth0Lock(clientId, domain, {
    plugins: [new CustomPlugin()],
});

lock.show();
```


luisrudge:
Thanks for the PR @lexaurin 🎉

luisrudge:
Hi @lexaurin can you see why the test is failing?

lexaurin:
Hi @luisrudge, sorry, missed getClient call when cherry picking from previous PR. Should be fixed now.

PR #1225, opened 2018-01-16
Update auth0.js

luisrudge:
Updating auth0.js with updated versions of idtoken-verifier and superagent and also https://github.com/auth0/auth0.js/pull/615
superagent was moved to a devDependency as well because it's not used anymore inside lock itself.


luisrudge:
I just fixed a previous written test that was snapshotting auth0.js version, which would fail on every upgrade.

Issue #1220, opened 2018-01-11
Additional flashMessage type: `info`


selaux:
Would you accept a PR which introduces a new `flashMessage` type? Currently afaik only `success` and `error` are possible. We want to have a third option `info`, which would show in blue. This way we could inform the user about things without having them thinking that something good/bad happened...

luisrudge:
What types of auth related things would you inform there?

selaux:
All sorts of auth related things 😉 . E.g.

- if the user just changed his email it should show him a message saying something like: "Please verify your new email before logging in"
- if the user just did a booking with us, it should show something like: "Please login to see your booking details" (or "... register ..." for non-registered users)

Basically we want to be able to show auth related messages without communicating them as "successful" or "error" messages.

luisrudge:
Hmm. I guess it makes sense. Ok, let's see that PR 🎉 If we don't have an 'info color', please use #44C7F4 as background color to follow our [styleguide](https://styleguide.auth0.com/#/design/status-colors)

fstoerkle:
We just opened a PR for this: https://github.com/auth0/lock/pull/1233

selaux:
PR was merged. Will close.

PR #1233, opened 2018-01-24
Add new flash message type `info`

fstoerkle:
As discussed in #1220, we want to add a new `flashMessage` type `info` in order to show messages that are neither an error nor a success message.

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -134,6 +134,18 @@ export function clearGlobalSuccess(m) {
   return tremove(m, 'globalSuccess');
 }
 
+export function setGlobalInfo(m, str) {
+  return tset(m, 'globalInfo', str);
+}
+
+export function globalInfo(m) {
+  return tget(m, 'globalInfo', '');
+}
+
+export function clearGlobalInfo(m) {
+  return tremove(m, 'globalInfo');
+}
+
 export function rendering(m) {
   return tget(m, 'render', false);
 }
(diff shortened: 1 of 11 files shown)


src/core/index.js
```diff
@@ -563,8 +575,9 @@ export function overrideOptions(m, opts) {
   }
 
   if (opts.flashMessage) {
-    const key = 'success' === opts.flashMessage.type ? 'globalSuccess' : 'globalError';
-    m = tset(m, key, opts.flashMessage.text);
+    const { type } = opts.flashMessage;
```
luisrudge: this will throw if flashMessage.type is undefined. use `error` as a default (+ a test)

luisrudge:
Thanks @fstoerkle. Just left a single comment on something I found. 🎉

src/core/index.js
fstoerkle: Makes sense.

However, I just had a look at the tests and there is one test case explicitly asserting that Lock should fail if there is no `type` set on a `flashMessage`:
https://github.com/auth0/lock/blob/master/test/show_with_flash_message.test.js#L19
This kind of contradicts your suggestion to make the `type` optional and falling back to `error` as default ` type`.

I'm fine with both options, what do you prefer?

luisrudge: You're right. I didn't remember that 😝 
Let's leave it like this. I'll test the PR today and merge it. Thanks! 🎉

fstoerkle: @luisrudge Hi Luis, did you manage to take a look at this again in the meanwhile?

luisrudge:
Ops, sorry! Looking at it now.

luisrudge:
Thanks for the PR 🎉 
It will be in the next release

fstoerkle:
🚀  Sweet, thanks for taking care of this!

PR #1239, opened 2018-01-25
Add support for html in the additional checkbox label

luisrudge:


src/__tests__/field/__snapshots__/custom_input.test.jsx.snap
```diff
@@ -11,9 +11,13 @@ exports[`CustomInput when type == checkbox renders correctly as a CheckBoxInput
       onChange={[Function]}
       type="checkbox"
     />
-    <span>
-      placeholder
-    </span>
+    <span
+      dangerouslySetInnerHTML={
+        Object {
+          "__html": "placeholder",
```
lbalmaceda: Trailing comma

luisrudge: this is a snapshot file. it is auto-generated

luisrudge:
No. You can't inject scripts:

```js
additionalSignUpFields: [{
    type: "checkbox",
    name: "newsletter",
    prefill: "true",
    placeholder: "I <script>alert('hi');<\/script>",
  }]
```

screen shot 2018-01-25 at 15 37 22

lbalmaceda:
Maybe not directly in the string code but you could be linking to a script.  

>In general, setting HTML from code is risky because it’s easy to inadvertently expose your users to a cross-site scripting (XSS) attack.

From oficial docs: https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

luisrudge:
You mean you could link to a malicious site? Sure. Every link can do that. Not much we can do in that regard. Since it's the only way to add raw html from a variable, we already use `dangerouslySetInnerHTML` in several places: https://github.com/auth0/lock/search?utf8=%E2%9C%93&q=dangerouslySetInnerHTML&type=
It could be an issue if this was something that was user-generated (comments, posts etc). In our case, it's not, so we're good

Issue #1261, opened 2018-02-10
v11 onAuthorizationError error description is undefined


rokadias:
I switched to v11 (11.2.3) of the lock and now my flashMessage that I send to the lock is undefined, because the description is no longer populated. This happens for invalid credentials.

```javascript
onAuthorizationError(error: any): void {
    var description = error.description || error.error_description;

    this.flashMessage = {
      type: 'error',
      text: description
    };
   this.lock.hide();
   this.lock.show({
      initialScreen:
        (description || '').includes("Please change your password")
          ? "forgotPassword"
          : "login",
      flashMessage: this.flashMessage
    });
};
```
I was using auth0-lock 10.22.0 and auth0-js 8.7.0. The code above works and when an invalid credential authorization failure occurs it's error.description. When my custom rule causes an unauthorizederror on the server it becomes error.error_description. From the network console I see a request to `/usernamepassword/login` with the following 
`{"name":"ValidationError","code":"invalid_user_password","description":"Wrong email or password.","statusCode":400}` You can see the error object contains `description`.

In v11 the request is made to `/co/authenticate` instead (I'm assuming because of the cross origin changes) and the response is: `{"error":"access_denied","error_description":"Wrong email or password."}`

However `error_description` is not included to onAuthorizationError. This is because of the code here:
https://github.com/auth0/lock/blob/fc16f1961b0bd30f509775e44206caa1536678c9/src/core/web_api/helper.js#L109-L115

There's also the matter of custom errors occuring on parseHash and that doesn't run through the same normalizeError function as the other login methods in p2_api.js.
https://github.com/auth0/lock/blob/fc16f1961b0bd30f509775e44206caa1536678c9/src/core/web_api/p2_api.js#L86-L94

For Example:
https://github.com/auth0/lock/blob/fc16f1961b0bd30f509775e44206caa1536678c9/src/core/web_api/p2_api.js#L82-L84

Which also doesn't help when there's a 3rd option of changing the description field name from error_description to `errorDescription` for authorization failures that occur in parseHash from the auth0.js
https://github.com/auth0/auth0.js/blob/8eaf976dc937ad36ca193297c30b117bec318f22/src/helper/error.js#L4

I'll make a PR to see if I can make this a bit more consistent, but this is what I'm trying to solve.

luisrudge:
Hi, I found the issue and it's inside auth0.js, so I'm closing this issue and your PR. This will be fixed in the next release of lock. Thanks for your patience!

[...]

PR #1262, opened 2018-02-10
Normalizing the error description field

rokadias:
I found a number of cases where the description returned either from an api end point or rewritten from auth0 js was causing the error object passed to onAuthorizationError to no longer contain the description.

This my attempt to fix it. To find out more read the issue: #1261


luisrudge:
Hi, thanks for the PR. I'll investigate this in the next week. [This code](https://github.com/auth0/auth0.js/blob/master/src/helper/response-handler.js#L47) of auth0.js should be handling the error mapping. :(

luisrudge:
fixed in https://github.com/auth0/auth0.js/pull/667

PR #1266, opened 2018-02-15
Removing legacy COA mapping + fixing access_denied mapping

luisrudge:


src/__tests__/core/web_api/__snapshots__/helper.test.js.snap
```diff
@@ -1,15 +1,14 @@
...
-exports[`normalizeError maps COA not enabled error 1`] = `
+exports[`normalizeError should map access_denied error to invalid_user_password when error.code === access_denied 1`] = `
 Object {
-  "code": "cross_origin_login_not_enabled",
-  "description": "Cross Origin Authencation is not enabled. Please refer to this document to enable it: https://link-to-doc.com",
-  "error": "cross_origin_login_not_enabled",
-  "logToConsole": true,
+  "code": "invalid_user_password",
+  "description": "foobar",
+  "error": "invalid_user_password",
```
lbalmaceda: What's the reason behind having code and error with the same values? Shouldn't you keep 1, `code` probably?

src/__tests__/core/web_api/helper.test.js
```diff
@@ -43,4 +34,17 @@ describe('normalizeError', () => {
...
   });
+  it('should map access_denied error to invalid_user_password when error.code === access_denied', () => {
+    const error = {
+      code: 'access_denied',
+      description: 'foobar'
+    };
+    const expectedError = {
+      code: 'invalid_user_password',
+      error: 'invalid_user_password',
+      description: error.description
```
lbalmaceda: use `description: 'foobar'` instead of error.description

src/__tests__/core/web_api/__snapshots__/helper.test.js.snap
luisrudge: I don't know why it is like this. Can't change this now though.

src/__tests__/core/web_api/helper.test.js
luisrudge: fixed

Issue #1250, opened 2018-02-05
Consider HRD configuration of enterprise connections in forgot password


jmangelo:
The available forgot password functionality can be confusing in the following scenario:

- there's a database and an enterprise connection enabled (the enterprise one has HRD configured for `@contoso.com` emails).
- an end-user inputs an SSO-enabled email address in the forgot password field.

The forgot password should detect when an enterprise connection email address is entered, and suggest that it isn't possible to reset that password via Lock. However, currently the end-user can continue with the reset password request even though that if the same email is entered in login in SSO will be forced.

Tested with Lock 10.24.3 and 11.2 version.

luisrudge:
related: ZD: 34486

PR #1278, opened 2018-02-28
Making HRD work in signup page and showing an error when HRD is detected in the forgot password page

luisrudge:
fix https://github.com/auth0/lock/issues/1250

image


joshcanhelp:
@lbalmaceda - I wrote that and agree that it doesn't look great as all-caps. We could remove that first sentence, TBH, and I think the message would be understandable. It's really just there to provide a path forward for folks. I don't think the end user is going to care about Enterprise or not.

luisrudge:
as for the translation: they probably weren't translated before.
as for the text, I don't think it's too big, in any case the customer can change that so, is that a big deal?

src/connection/database/reset_password.jsx
```diff
@@ -36,6 +40,21 @@ export default class ResetPassword extends Screen {
   getScreenTitle(m) {
     return i18n.str(m, 'forgotPasswordTitle');
   }
+  isSubmitDisabled(m) {
```
lbalmaceda: Where is this used?

Issue #1180, opened 2017-11-27
Whitespace not being trimmed for email input when username_required


MrPrashantT:
If connection does not require username, entering email with trailing spaces authenticates successfully (Lock trims the whitespace). If connection requires username, Lock doesn't trim the whitespace, causing auth to fail.

Repro on Lock v10.20.0:

1. Enable `requires username` on connection associated with client
2. Attempt to loginon Lock with trailing whitespace on email input `prashant@auth0.com ` 
3. Auth will fail. Monitoring network call, can see that whitespace is not trimmed:

```
//request payload to usernamepassword/login

{
  ...
  "connection": "Username-Password-Authentication",
  "username": "prashant@auth0.com ",
  "password": "<password>",
  ...
}
```

The same request, disabling `requires username` in the connection settings, results in Lock trimming the whitespace and successful auth.

e1senh0rn:
Any updates here?

luisrudge:
Hi. Not yet. We have other priorities right now. Do you mind sending a PR? I can review it and ship it if that's the case

brassier:
This is also a fairly large issue for us.  This was causing lots of our users to be "blocked", when they really were not blocked.  Invalid users were being reported as blocked (e.g. `blah@gmail.com<SPACE>`), confusing lots of users and support staff.

We really need this fixed before we can upgrade to 11.x.

brassier:
@luisrudge - I spent some time trying to dig through the Auth0 layers in GitHub.  It looks to me like this may boil down to an Auth0.js change and not Lock.  Does that sound right to you?  If this is correct, would we need to wait for a change in that library, and then a change for Lock to pull in that new version?  We need this fixed so we can upgrade to Lock v11 prior to the deprecation in April.

Add trim here:
https://github.com/auth0/auth0.js/blob/master/src/web-auth/username-password.js#L21

And possibly here:
https://github.com/auth0/auth0.js/blob/master/src/web-auth/cross-origin-authentication.js#L48

See (very old) auth.js library that used to trim similar code:
https://github.com/auth0/auth0.js/blob/038945972264ec4ef8dae5ca055f82f6fda0d888/index.js
`username:   trim(options.username || options.email || ''),`

PR #1280, opened 2018-03-01
Always remove spaces from email and username

luisrudge:
fix https://github.com/auth0/lock/issues/1180


src/__tests__/field/email.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+    jest.mock('field/index', () => ({
+      setField: jest.fn()
+    }));
+
+    email = require('field/email');
+  });
+  describe('setEmail()', () => {
+    it(`removes spaces`, () => {
+      email.setEmail(Immutable.fromJS({}), ' email@te  st.com ');
```
lbalmaceda: While I agree leading and trailing whitespaces should be trimmed, i think having whitespaces on the middle of an email address or username is an "invalid value" rather.

luisrudge: why show "invalid field" if we can remove them at once and move on?

joshcanhelp: Was just typing this out, actually. Fully agree (if spaces are actually invalid characters ... are they excluded by the platform?)

lbalmaceda: Because it's a user's typing error and we should point them that it's an invalid value and they should fix it. If we don't, they might think they registered using "this email @address.com" and when they try to log in on other platform they won't be allowed.

Also, A0 username regex: `^[a-zA-Z0-9_]+$`

Screenshot below from gmail when typing an email address:
image

luisrudge: fixed

src/field/email.js
```diff
@@ -15,8 +15,8 @@ export function isEmail(str) {
   return !!result && result[0] !== null;
 }
 
-export function setEmail(m, str) {
-  return setField(m, 'email', str, str => {
+export function setEmail(m, str = '') {
```
lbalmaceda: you're adding a default value here (if str is not present). Isn't this breaking further validation?

luisrudge: I just wanted to be sure you would not get null/undefined, but the code is already not throwing [here](https://github.com/auth0/lock/blob/bb92858006eb1b39e38a9fb9000704554dc69997/src/field/email.js#L14), so we're fine.

Issue #1235, opened 2018-01-24
v11 not working in cordova


lexaurin:
After upgrade from version 10 to version 11 we have problem logging in with username/password method on Cordova environment. In a regular browser login works.

Logging in Cordova with username/password ends up with "WE'RE SORRY, SOMETHING WENT WRONG" message.

Our configuration is:
```
new Auth0Lock(clientId, domain, {
                    autoclose: true,
                    auth: {
                        sso: false,
                        redirect: true, // Automatically FALSE in Cordova
                        redirectUrl: "http://192.168.0.12:9090/api/v1/auth/oauth/auth0",
                        responseType: 'code'
                    }});
```

What I've figured out so far:
1) [origin validation added in latest 11.1.0](https://github.com/auth0/auth0.js/commit/ce402c8628c8f34a69145415fecbb5233a9500d3#diff-b1ca239731d0a93b9f99651b387bfa4bR40) fails on Cordova. On desktop it's ok because in redirect mode there is no such check. While in cordova it falls to popup mode (which is ok) but then origin check fails. `currentOrigin` is `"file://"` but `redirectUriOrigin` is `undefined`. `options` object doesn't contain `redirectUri` prop. 
![screen shot 2018-01-24 at 18 30 26](https://user-images.githubusercontent.com/428302/35347234-cac9b6a4-0134-11e8-9eb5-ecc637ca3b7b.png)

2) When we downgraded to 11.0.1 to fix first mentioned issue, it doesn't work because of cross-origin check. Our page is on `file://` protocol, such as `file:///data/content/tmp/e8f41568` but it try to load invisible iframe with auth0.com domain. So we have to add our domain to "Allowed Web Origins" I presume but  I can't because `file://` is not valid origin in Auth0 Dashboard form. It does not have <host> part. 
![screen shot 2018-01-24 at 18 34 09](https://user-images.githubusercontent.com/428302/35347397-4c98406a-0135-11e8-940d-e1ae79323c80.png)

Any thoughts?

luisrudge:
I'm fixing the origin check right now: https://github.com/auth0/auth0.js/pull/635 
should be released today. Do you think that will fix though?

lexaurin:
Hi Luis, 

thanks for the fix, just tried it out on Cordova but sadly it doesn't fix issue 1. nor issue 2. 

[...]

PR #1283, opened 2018-03-05
Backward compatibility for name/pass login on cordova.

lexaurin:
Fix 1) for https://github.com/auth0/lock/issues/1235
Adds similar check for cordova environment as in old legacy_api and fires loginWithResourceOwner() to login.


luisrudge:
Lock v11 is not going to work in Cordova apps, so you will need to migrate to the Hosted Login Page. More info here: https://github.com/auth0/lock/issues/1235#issuecomment-374740654

PR #1289, opened 2018-03-08
Fix Romanian translations for Auth0 lock login dialog

paulborza:
The translation problem can be seen on the Image I uploaded [here](https://imgur.com/a/nCk8g)


luisrudge:
Thanks for the PR 🎉

PR #1290, opened 2018-03-09
Update auth0.js and add flag to enable impersonation

luisrudge:


package.json
```diff
@@ -87,7 +87,7 @@
     "zuul-ngrok": "4.0.0"
   },
   "dependencies": {
-    "auth0-js": "^9.3.0",
+    "auth0-js": "^9.3.3",
     "blueimp-md5": "2.3.1",
     "fbjs": "^0.3.1",
     "idtoken-verifier": "^1.0.1",
```
lbalmaceda: Shouldn't this be bumped too?

src/__tests__/core/web_api/p2_api.test.js
```diff
@@ -159,4 +159,22 @@ describe('Auth0APIClient', () => {
...
+    it('should pass __enableImpersonation when options._enableImpersonation===true', () => {
+      const client = getClient({ _enableImpersonation: true });
+      client.parseHash('hash', 'cb');
+      const mock = getAuth0ClientMock();
+      const parseHashMock = mock.WebAuth.mock.instances[0].parseHash.mock;
+      expect(parseHashMock.calls.length).toBe(1);
+      expect(parseHashMock.calls[0]).toMatchSnapshot();
+    });
+  });
 });
```
lbalmaceda: Besides checking that the option is set on the auth0js client, you can also check that a real auth0js client can parse or throws an exception depending on the flag and on the `state` passed in the hash. (2 more tests)

package.json
luisrudge: I removed the direct dependency

src/__tests__/core/web_api/p2_api.test.js
luisrudge: I'm not testing auth0.js internals inside lock

PR #1291, opened 2018-03-10
add arabic support lang

hamzahio:


luisrudge:
you removed the en.js file, can you put it back? Thanks for the PR 🎉

PR #1304, opened 2018-03-21
Added .babelrc to .npmignore

kedrzu:
Your `.babelrc` file is published to npm. This causes i.e that if I want to transpile my project along with dependencies, I am dependent on your babel configuration (plugins, presets, etc).

Could you please add `.babelrc` file to `.npmignore`?

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -13,3 +13,4 @@ bower.json
 Gruntfile.js
 .css.map
 *~
+.babelrc


luisrudge:
can you point me to other projects that does this? I'm not sure what's the guideline for this file tbh

kedrzu:
You can look i.e. on `vuejs` or `react`. They, however, do not use `.npmignore`, because they have a quite customized build chain, so they don't need it. They both use `babel`, but if you look into their npm packages, there will be no `.babelrc` file. 

For a full disclosure, why excluding `.babelrc` from npm would help:

Let's say you have a project with some third-party dependencies written in ES6. To be ES5 compliant you would use babel to transpile your project as well as your dependencies. When you include i.e. `babel-loader` for `node_modules` dependencies it would do the job.

Babel will then traverse directory tree to search for .babelrc file. Because `auth0-lock` includes its own `.babelrc` in npm package, the result is that it is being used by babel, along with your configured presets and plugins.

This results in indirect dependencies of my project to your babel setup and the need to install all the presets and plugins you use.

luisrudge:
Thanks for the explanation. That makes a lot of sense. I also found this: https://github.com/facebook/react-native/issues/4062#issuecomment-155662710

kedrzu:
Great, you mentioned it, there is plenty of other npm modules that include `.babelrc`. I would have better argument :)

PR #1307, opened 2018-03-22
Fixed German translations

roschaefer:
* I used the formal way of addressing people, i.e. "Sie" instead of colloquial "Du"
* I am not able to see the effect of breaks inside the text (`<br>`), if it looks OK etc.


luisrudge:
Thanks for the PR 🎉

PR #1309, opened 2018-03-22
Fix inconsistent state assignment

luisrudge:
The correct way of sending the `state` parameter in the authentication request is

```js
var options = {
  auth: {
    params: { state: 'foobar' }
  }
}
```

But, because of the bug I just fixed, this state, although being sent in the auth requests, wasn't being sent in the parseHash method, which caused the '`state` doesn't match' error.
This gets the correct state without breaking anyone.

With this code, both version will work for parsing the hash with a specific state:

```js
var options = {
  auth: {
    params: { state: 'foobar' }
  }
}
```

```js
var options = {
  auth: {
    state: 'foobar'
  }
}
```

### DIFF
--- a/src/core/web_api/p2_api.js
+++ b/src/core/web_api/p2_api.js
@@ -34,11 +34,16 @@ class Auth0APIClient {
       _telemetryInfo: opts._telemetryInfo || default_telemetry
     });
 
+    var state = opts.state;
+    if (opts.params && opts.params.state) {
+      state = opts.params.state;
+    }
+
     this.authOpt = {
       popup: !opts.redirect,
       popupOptions: opts.popupOptions,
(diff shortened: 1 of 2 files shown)


src/core/web_api/p2_api.js
```diff
@@ -34,11 +34,16 @@ class Auth0APIClient {
       _telemetryInfo: opts._telemetryInfo || default_telemetry
     });
 
+    var state = opts.state;
+    if (opts.params && opts.params.state) {
+      state = opts.params.state;
```
joshcanhelp: @luisrudge - if they have both then the former example in your description will override the latter. Is that expected? Is it consistently documented one way or the other?

Issue #1331, opened 2018-04-04
custom nonce not being sent with Lock v11 in IE10 with SSO


Pargus:
Lock v11 fails to initialize in IE10 because our custom nonce value is not included in the initial /authorize request that is being made by Lock (we need to provide our own nonce value because IE10 does not support the Web Crypto API). The initial /authorize request is being made by Lock because we have auth.sso set to true so that the last login is remembered for the user. It looks like a bug that the custom nonce value is not included in this request but please enlighten me if there is any valid reason for this. If I set auth.sso to false everything works fine but we want the last login to be remembered so it has to be true.

When I debug lock.js (v11.5.2) I can see that the initial `checkSession` call (inside function `Authentication.prototype.getSSOData`) is made without the nonce property:
![image](https://user-images.githubusercontent.com/4414762/38315698-4b95dad0-3829-11e8-8607-e578fd7338ec.png)
which forces a nonce to be generated in `TransactionManager.generateTransaction` using the `randomString` function:
![image](https://user-images.githubusercontent.com/4414762/38318602-0275f586-3830-11e8-82d9-f9a9ba406484.png)
But in IE10 the `crypto`/`msCrypto` object is not present so null is returned which causes the Lock widget to get stuck with the spinning loader and logs the error "nonce option is required". 

I can see in Chrome that an auto generated nonce is used for that initial request but for all subsequent requests my custom nonce (in auth.params.nonce) is indeed included. Can't the nonce just be grabbed from auth.nonce for that initial sso authorize request, just like the auth.redirectUrl value is? I am not sure where things need to change in lock.js for this to happen.

Thanks,
Pär

luisrudge:
how are you initializing lock?

maxlapides:
I am experiencing the same issue. You can reproduce this issue by loading any instance of Auth0 Lock in IE 10. I recommend browserstack.com for testing.

In the meantime, here's a hack you can use to get Auth0 Lock to load in IE 10. But I would like to see a resolution to this issue as soon as possible since I'm certain this is not as secure as it should be.

```javascript
if (!window.crypto && !window.msCrypto) {
  window.crypto = {
    getRandomValues: array => {
      for (let i = 0; i < array.length; i++) array[i] = _.random(0, 64)
      return array
    }
  }
}
```

Pargus:
@luisrudge I'm initializing Lock like this:
```js
[...]

PR #1333, opened 2018-04-05
Fix getSSOData not sending custom nonce/state

luisrudge:
fix https://github.com/auth0/lock/issues/1331

### DIFF
--- a/README.md
+++ b/README.md
@@ -520,7 +520,7 @@ var lock = new Auth0Lock(clientId, domain, options);
 lock.show();
 ```
 
-More information can be found in [Auth0's documentation](https://auth0.com/docs/libraries/lock/popup-mode).
+More information can be found in [Auth0's documentation](https://auth0.com/docs/libraries/lock/v11/authentication-modes#popup-mode).
 
 ## Browser Compatibility
 


src/__tests__/core/web_api/p2_api.test.js
```diff
@@ -79,6 +83,14 @@ describe('Auth0APIClient', () => {
...
         expect(client.authOpt.state).toBe('foo');
       });
+      it('should set nonce from options.nonce', () => {
+        const client = getClient({ nonce: 'foo' });
+        expect(client.authOpt.nonce).toBe('foo');
+      });
+      it('should set nonce from options.params.nonce', () => {
+        const client = getClient({ params: { nonce: 'foo' } });
+        expect(client.authOpt.nonce).toBe('foo');
+      });
```
lbalmaceda: what if both `nonce` are set, which one should win? Add that test and document this behavior somewhere.

src/__tests__/core/web_api/p2_api.test.js
```diff
@@ -79,6 +83,14 @@ describe('Auth0APIClient', () => {
         const client = getClient({ params: { state: 'foo' } });
         expect(client.authOpt.state).toBe('foo');
       });
```
lbalmaceda: what if both `state` are set, which one should win? Add that test and document this behavior somewhere.

luisrudge: added

src/__tests__/core/web_api/p2_api.test.js
luisrudge: added

PR #1339, opened 2018-04-10
Allow select field arrow to be clicked

nmchaves:
Currently, if a user clicks directly on the arrow icon in a select field, nothing will happen.

This PR applies `pointer-events: none` to the arrow icon. As a result, clicking on the icon will now open the select menu.


luisrudge:
Thanks for the PR 🎉

nmchaves:
Thanks @luisrudge!

PR #1344, opened 2018-04-16
Introduce logo link theme option

SergeL:
Since it's a common use case to link the logo to the copmpany's home or other page we want to suggest a new theme option: `logoLink` in this PR.

This option will make the logo in the Lock header a link to the specified URL.

It makes theming and customising the auth0-lock ui component more flexible 💪.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,7 @@ var options = {
   theme: {
     labeledSubmitButton: false,
     logo: "https://example.com/assets/logo.png",
+    logoLink: "https://example.com/",
     primaryColor: "green",
     authButtons: {
       connectionName: {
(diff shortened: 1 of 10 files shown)


luisrudge:
I'm talking with the team internally about this.. Meanwhile, can you please fix the test? Thanks for the PR 🎉

SergeL:
@luisrudge thanks for the feedback.

The test should be fixed now 👍

lo1tuma:
Are there any news regarding this feature?

luisrudge:
Hi, sorry I forgot to come back with our team's feedback. We decided this is not something we want inside Lock, considering we don't want users to accidentally click somewhere and then get redirected away from the authentication process. Thanks for the PR anyway 🎉

lo1tuma:
Thanks for the update. Although I don’t quite understand the reasoning behind this decision.

* This feature is implemented behind an option, which is disabled per default. So nobody is affected by this when this feature is not explicitly enabled.
* What if a user navigates to the login page and then decides to **not** login? Currently there is no way for the user to get back.
*  There are still ways to achieve this, but I would consider alternative solutions only as workaround because the require quite a lot of code to implement. For example we have this feature currently implemented on our hosted page by registering a global click listener for the logo element and then we redirect by setting `window.location.href`. Another solution would be to completely stop using the the "logo feature" from `lock` and to display to logo separately (outside of `lock`) which can then be linked to any page.

luisrudge:
> What if a user navigates to the login page and then decides to not login? Currently there is no way for the user to get back.

This was one of the points made by our team. The user is probably coming from the main website to the hosted login page, so there's no point in linking to the main website again. If the user wants to go back, they just need to hit the back button in the browser.

PR #1350, opened 2018-04-17
Update en.js

senorcodecat:
Typo for enterpriseLoginInstructions. Missing an 's'.


luisrudge:
ouch! Although there's clearly a typo here, we can't merge this PR because it'd break everyone overriding this entry. To be able to merge this, you also have to update all language files and usage of the `enterpriseLoginIntructions` entry across the repo, making it use `enterpriseLoginInstructions` first and if the string is empty, fallback to `enterpriseLoginIntructions`.

PR #1368, opened 2018-05-03
Fix password managers submitting the form incorrectly

luisrudge:
Discussion about the issue: https://spectrum.chat/?t=90bc0f1b-54fc-40f4-b718-353e80b47cdc

TLDR: Password managers use `form.submit()` to submit the form. This, according to the [docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit), prevents the `onsubmit` event from being emitted, which prevents react to handle the event.

> When invoking this method directly, no submit event is raised (in particular, the form's onsubmit event handler is not run),  and constraint validation is not triggered either.

The fix is also in the docs: 

> If a form control (such as a submit button) has a name or id of submit it will mask the form's submit method.

So, all we have to do is to add `name="submit"` to our submit button. In our specific case, I also had to change the behavior to make sure the button is always inside the dom (it was being added/removed  based on specific properties). Now, it's always in the DOM, but we toggle its visibility with a custom `style` rule.

### DIFF
--- a/src/ui/box/chrome.jsx
+++ b/src/ui/box/chrome.jsx
@@ -305,6 +305,7 @@ export default class Chrome extends React.Component {
             which doesn't trigger the `onsubmit` event handler, which
             makes impossible to react to handle the submit event, 
             causing the page to send a POST request to `window.location.href`
+            with all the form data.
          */}
         <SubmitButton
           color={primaryColor}
--- a/src/ui/box/chrome.jsx
+++ b/src/ui/box/chrome.jsx
@@ -303,7 +303,7 @@ export default class Chrome extends React.Component {
             The submit button should always be included in the DOM.
             Otherwise, password managers will call `form.submit()`,
             which doesn't trigger the `onsubmit` event handler, which
-            makes impossible to react to handle the submit event, 
+            makes impossible for react to handle the submit event, 
             causing the page to send a POST request to `window.location.href`
             with all the form data.
          */}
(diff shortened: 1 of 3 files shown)


src/ui/box/chrome.jsx
```diff
@@ -309,7 +299,23 @@ export default class Chrome extends React.Component {
             </div>
           </MultisizeSlide>
         </div>
-        {submitButton}
+        {/*
+            The submit button should always be included in the DOM.
+            Otherwise, password managers will call `form.submit()`,
+            which doesn't trigger the `onsubmit` event handler, which
+            makes impossible to react to handle the submit event,
```
lbalmaceda: for react*

src/ui/box/chrome.jsx
```diff
@@ -309,7 +299,23 @@ export default class Chrome extends React.Component {
             </div>
           </MultisizeSlide>
         </div>
-        {submitButton}
+        {/*
```
lbalmaceda: previous code wrapped the `<SubmitButton.../>` in between `{}`. Now you're just wrapping a comment and then adding the button. Is that OK?

src/ui/box/chrome.jsx
```diff
@@ -309,7 +299,23 @@ export default class Chrome extends React.Component {
...
+            causing the page to send a POST request to `window.location.href`
+            with all the form data.
+         */}
+        <SubmitButton
+          color={primaryColor}
+          disabled={disableSubmitButton}
+          screenName={screenName}
+          contentProps={contentProps}
+          label={submitButtonLabel}
+          ref={el => (this.submitButton = el)}
```
lbalmaceda: why is the `key` parameter gone?

PR #1370, opened 2018-05-06
Supports ShadowDom in `container` option

### LINKED ISSUE #1369: `container` option does not support shadowDom
Thank you for using Lock! Please read the following before creating a new issue.

- [ ] Is this a question about Auth0 in general or a support request? Get in touch with us at https://support.auth0.com.
- [x] Is this issue something you can fix? Please send us a pull request! Bug fixes are welcome.

Lock does not allow passing a container element which is obscured behind a shadowDom boundary.

```js
const show = () => {
  const lock = new Auth0LockPasswordless(clientID, domain, { auth, theme });
  lock.show(clientID, domain, {
    container: document.getElementById('app').shadowRoot.getElementById('login')
  });
};
```

Instead, it only accepts a string for id in the light dom.

https://github.com/auth0/lock/blob/b66b79917d66b76184e835fa760c80f40ee4aebb/src/ui/box.js#L8

This property should check the type passed, and if it's an element, accept it as is.

bennypowers:
This PR (basically a straw man unless it works ;D) allows devs to pass a handle on an element as `container` to lock.show, which is good enough to mount the component on an element behind a shadow boundary.

closes #1369


luisrudge:
I like this! Can you please fix the `const` error and add a test? 🎉

luisrudge:
I see a lot of errors in the tests. where is `isModal` coming from?

bennypowers:
Gotta be honest with you I'm not familiar enough with the codebase to say what's really going on here. maybe @gnandretta could chime in?

luisrudge:
I know `isModal` comes from here: https://github.com/auth0/lock/blob/21865a5605685da2ac6fdf8294ca1a7c8b93733d/src/core.js#L120, but your code is trying to access the variable without declaring it:

https://github.com/auth0/lock/pull/1370/files#diff-0c361d252ff34c87addf0b44479c04c5R55
https://github.com/auth0/lock/pull/1370/files#diff-0c361d252ff34c87addf0b44479c04c5R75

PR #1373, opened 2018-05-10
Fix IE default redirect url

### LINKED ISSUE #1361: JS Console Error - 'unable to get property 'protocol' of undefined or null reference' on Win8-IE10 and Win 8.1-IE11
Hey.
We are getting a JS error 'unable to get property 'protocol' of undefined or null reference'

Browser - IE 10 (Win 8) and IE 11 (Win 8.1)

This happened on upgrading Auth0 Lock from 10.9.1 to 11.3.0 and Auth0 from 8.0.4 to 9.3.3!

screen shot 2018-03-22 at 8 34 24 pm

On using debugger we found that Line no. 4171 in lock.min.js is the cause. Following code snippet suggests the same, where e is undefined:

`function d(t) {
        if(t) {
                var e = g(t), n = e.protocol + "//" + e.hostname;
                return e.port && (n += ":" + e.port) , n;
        }
 }`

Also I am attaching lock.min.js (renamed as lock.min.txt here) copied from Win 8.1 IE 11 browser debugger. 
[lock.min.txt](https://github.com/auth0/lock/files/1963225/lock.min.txt)

This is how we have included the libraries.
<script src="https://cdn.auth0.com/js/lock/11.3.0/lock.min.js"></script>
<script src="https://cdn.auth0.com/js/auth0/9.3.3/auth0.min.js"></script>

luisrudge:
fix #1361

### DIFF
--- a/src/__tests__/core/index.test.js
+++ b/src/__tests__/core/index.test.js
@@ -37,6 +37,15 @@ describe('setup', () => {
     const model = mock.calls[0][1].toJS();
     expect(model.auth.redirectUrl).toBe('https://test.com/path/');
   });
+  it('default redirectUrl should work when `window.location.origin` is not available', () => {
+    setURL('https://test.com/path/#not-this-part', { noOrigin: true });
+    const options = {};
+    setup('id', 'clientID', 'domain', options, 'hookRunner', 'emitEventFn');
+    const { mock } = mockInit;
+    expect(mock.calls.length).toBe(1);
+    const model = mock.calls[0][1].toJS();
+    expect(model.auth.redirectUrl).toBe('https://test.com/path/');
+  });
   it('should work with redirect:false and responseType:id_token', () => {
     const options = {
       auth: {
--- a/src/__tests__/utils/url_utils.test.js
+++ b/src/__tests__/utils/url_utils.test.js
@@ -7,17 +7,34 @@ describe('url utils', () => {
       expect(getOriginFromUrl('')).toBe(undefined);
       expect(getOriginFromUrl(null)).toBe(undefined);
     });
-    it('should use an anchor to parse the url and return the origin', function() {
+    it('should parse the url and return the origin with https', function() {
       var url = 'https://test.com/example';
       expect(getOriginFromUrl(url)).toBe('https://test.com');
     });
-    it('should use add the `port` when available', function() {
+    it('should parse the url and return the origin with http', function() {
+      var url = 'http://test.com/example';
+      expect(getOriginFromUrl(url)).toBe('http://test.com');
+    });
+    it('should include the `port` when available', function() {
       var url = 'https://localhost:3000/example';
       expect(getOriginFromUrl(url)).toBe('https://localhost:3000');
     });
   });
   describe('getLocationFromUrl', function() {
+    it('should return null for invalid urls', () => {
+      expect(getLocationFromUrl('banana')).toBe(null);
+    });
     const mapping = {
+      'http://localhost:3000/foo?id=1': {
+        href: 'http://localhost:3000/foo?id=1',
+        protocol: 'http:',
+        host: 'localhost:3000',
+        hostname: 'localhost',
+        port: '3000',
+        pathname: '/foo',
+        search: '?id=1',
+        hash: ''
+      },
       'https://localhost:3000/foo?id=1': {
         href: 'https://localhost:3000/foo?id=1',
         protocol: 'https:',
--- a/src/__tests__/utils/url_utils.test.js
+++ b/src/__tests__/utils/url_utils.test.js
@@ -55,15 +55,25 @@ describe('url utils', () => {
         search: '',
         hash: ''
       },
-      'https://auth0.com#access_token=foo': {
-        href: 'https://auth0.com#access_token=foo',
+      'https://auth0.com/#access_token=foo': {
+        href: 'https://auth0.com/#access_token=foo',
         protocol: 'https:',
         host: 'auth0.com',
         hostname: 'auth0.com',
         port: undefined,
-        pathname: '',
+        pathname: '/',
         search: '',
         hash: '#access_token=foo'
+      },
+      'https://auth0.com/#/foo/access_token=foo': {
+        href: 'https://auth0.com/#/foo/access_token=foo',
+        protocol: 'https:',
+        host: 'auth0.com',
+        hostname: 'auth0.com',
+        port: undefined,
+        pathname: '/',
+        search: '',
+        hash: '#/foo/access_token=foo'
       }
     };
     for (const url in mapping) {
--- a/src/__tests__/utils/url_utils.test.js
+++ b/src/__tests__/utils/url_utils.test.js
@@ -1,0 +1,58 @@
+import { getOriginFromUrl, getLocationFromUrl } from '../../utils/url_utils';
+
+describe('url utils', () => {
+  describe('getOriginFromUrl', function() {
+    it('should return undefined if there is no url', function() {
+      expect(getOriginFromUrl()).toBe(undefined);
+      expect(getOriginFromUrl('')).toBe(undefined);
+      expect(getOriginFromUrl(null)).toBe(undefined);
+    });
+    it('should use an anchor to parse the url and return the origin', function() {
+      var url = 'https://test.com/example';
+      expect(getOriginFromUrl(url)).toBe('https://test.com');
+    });
+    it('should use add the `port` when available', function() {
+      var url = 'https://localhost:3000/example';
+      expect(getOriginFromUrl(url)).toBe('https://localhost:3000');
+    });
+  });
+  describe('getLocationFromUrl', function() {
+    const mapping = {
+      'https://localhost:3000/foo?id=1': {
+        href: 'https://localhost:3000/foo?id=1',
+        protocol: 'https:',
+        host: 'localhost:3000',
+        hostname: 'localhost',
+        port: '3000',
+        pathname: '/foo',
+        search: '?id=1',
+        hash: ''
+      },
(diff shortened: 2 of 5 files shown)


src/__tests__/core/index.test.js
```diff
@@ -37,6 +37,15 @@ describe('setup', () => {
     const model = mock.calls[0][1].toJS();
     expect(model.auth.redirectUrl).toBe('https://test.com/path/');
   });
+  it('default redirectUrl should work when `window.location.origin` is not available', () => {
+    setURL('https://test.com/path/#not-this-part', { noOrigin: true });
+    const options = {};
```
lbalmaceda: any reason why you have the declaration separate from the usage, since you're hardcoding the rest of the parameters too?

src/__tests__/utils/url_utils.test.js
```diff
@@ -0,0 +1,58 @@
...
+    it('should return undefined if there is no url', function() {
+      expect(getOriginFromUrl()).toBe(undefined);
+      expect(getOriginFromUrl('')).toBe(undefined);
+      expect(getOriginFromUrl(null)).toBe(undefined);
+    });
+    it('should use an anchor to parse the url and return the origin', function() {
+      var url = 'https://test.com/example';
+      expect(getOriginFromUrl(url)).toBe('https://test.com');
+    });
+    it('should use add the `port` when available', function() {
```
lbalmaceda: should ~use~ add

src/__tests__/utils/url_utils.test.js
```diff
@@ -0,0 +1,58 @@
...
+      var url = 'https://localhost:3000/example';
+      expect(getOriginFromUrl(url)).toBe('https://localhost:3000');
+    });
+  });
+  describe('getLocationFromUrl', function() {
+    const mapping = {
+      'https://localhost:3000/foo?id=1': {
+        href: 'https://localhost:3000/foo?id=1',
+        protocol: 'https:',
```
lbalmaceda: protocol should not include the `:`.

src/__tests__/utils/url_utils.test.js
```diff
@@ -0,0 +1,58 @@
...
+        pathname: '/foo',
+        search: '?id=1',
+        hash: ''
+      },
+      'https://auth0.com/foo': {
+        href: 'https://auth0.com/foo',
+        protocol: 'https:',
+        host: 'auth0.com',
+        hostname: 'auth0.com',
+        port: undefined,
```
lbalmaceda: Should you return `undefined` or `80` (default)?

src/__tests__/utils/url_utils.test.js
```diff
@@ -0,0 +1,58 @@
...
+        pathname: '/foo',
+        search: '',
+        hash: ''
+      },
+      'https://auth0.com#access_token=foo': {
```
lbalmaceda: If I enter manage.auth0.com, the `#` is put after the `/`.  .e.g.: `https://manage.auth0.com/#/applications`

src/__tests__/utils/url_utils.test.js
```diff
@@ -0,0 +1,58 @@
...
+        href: 'https://auth0.com#access_token=foo',
+        protocol: 'https:',
+        host: 'auth0.com',
+        hostname: 'auth0.com',
+        port: undefined,
+        pathname: '',
+        search: '',
+        hash: '#access_token=foo'
```
lbalmaceda: why does hash includes the `#`? Same as port property, it shouldn't. It's already implied in the property name 😂

src/core/index.js
```diff
@@ -273,7 +273,8 @@ function extractAuthOptions(options) {
   // if responseType was not set and there is a redirectUrl, it defaults to code. Otherwise token.
   responseType = typeof responseType === 'string' ? responseType : redirectUrl ? 'code' : 'token';
   // now we set the default because we already did the validation
-  redirectUrl = redirectUrl || `${window.location.origin}${window.location.pathname}`;
+  redirectUrl =
+    redirectUrl || `${getOriginFromUrl(window.location.href)}${window.location.pathname}`;
```
lbalmaceda: does this alternative of not using `window.location.origin` work in every scenario?

src/__tests__/utils/url_utils.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+describe('url utils', () => {
+  describe('getOriginFromUrl', function() {
+    it('should return undefined if there is no url', function() {
+      expect(getOriginFromUrl()).toBe(undefined);
+      expect(getOriginFromUrl('')).toBe(undefined);
+      expect(getOriginFromUrl(null)).toBe(undefined);
+    });
+    it('should use an anchor to parse the url and return the origin', function() {
+      var url = 'https://test.com/example';
```
lbalmaceda: what about `http` URLs?

src/utils/url_utils.js
```diff
@@ -1,5 +1,29 @@
-export function parseUrl(str) {
```
joshcanhelp: Is it breaking to remove this?

src/__tests__/utils/url_utils.test.js
```diff
@@ -0,0 +1,58 @@
...
+      expect(getOriginFromUrl(url)).toBe('https://test.com');
+    });
+    it('should use add the `port` when available', function() {
+      var url = 'https://localhost:3000/example';
+      expect(getOriginFromUrl(url)).toBe('https://localhost:3000');
+    });
+  });
+  describe('getLocationFromUrl', function() {
+    const mapping = {
```
joshcanhelp: Might want to add a subdomain test here

src/__tests__/utils/url_utils.test.js
joshcanhelp: Actually, the function does return the protocol with a colon:

```
// http://banana.com
host:"banana.com"
hostname:"banana.com"
href:"http://banana.com"
protocol:"http:"
```

[Wikipedia says](https://en.wikipedia.org/wiki/URL#Syntax) that protocol is:

> A non-empty scheme component followed by a colon

... which tells me that protocol should not include it. That said, type this into your console:

```
> window.location
...
< protocol: "https:"
```

The built-in `URL` object (MDN) has the colon so I think this is the right implementation. Again, in the console:

```
> var url = new URL(window.location.toString()); url;
...
< protocol: "https:"
```

src/__tests__/utils/url_utils.test.js
joshcanhelp: ~~use add~~ include

src/__tests__/utils/url_utils.test.js
joshcanhelp: It's text parsing, silly! No port in the text === no port found

src/__tests__/utils/url_utils.test.js
joshcanhelp: Same as `protocol` above, browser parsing with `URL` includes the `#`

src/utils/url_utils.js
```diff
@@ -1,5 +1,29 @@
...
+      port: match[4],
+      pathname: match[5],
+      search: match[6],
+      hash: match[7]
+    }
+  );
+}
+
+export function getOriginFromUrl(url) {
+  if (!url) {
```
joshcanhelp: You're catching an empty string here but ... 👇

src/utils/url_utils.js
```diff
@@ -1,5 +1,29 @@
...
+    return undefined;
+  }
+  var parsed = getLocationFromUrl(url);
+  var origin = parsed.protocol + '//' + parsed.hostname;
```
joshcanhelp: ... this is also a source of an error if you just pass in a string:

```
> getLocationFromUrl('banana')
< null
```

If there is no match then `null` is returned.

src/core/index.js
luisrudge: yes, if anything, it's way more reliable, since works across all browsers.

src/__tests__/utils/url_utils.test.js
luisrudge: yeah, it's what `window.location.hash` returns

src/__tests__/utils/url_utils.test.js
luisrudge: /# and # is the same thing

src/__tests__/utils/url_utils.test.js
luisrudge: lucho is discovering all the greatness or javascript. our boy is growing :tada

src/__tests__/core/index.test.js
luisrudge: easy to copy/paste later if needed :grimacing:

src/__tests__/utils/url_utils.test.js
luisrudge: pushed some tests with http

src/utils/url_utils.js
luisrudge: nope. internal use only

src/utils/url_utils.js
luisrudge: good catch. pushed a fix.

src/__tests__/utils/url_utils.test.js
lbalmaceda: I tried:
- https://manage.auth0.com/#/applications -> fine
- https://manage.auth0.com#/applications -> redirs to https://manage.auth0.com/#/applications
- https://manage.auth0.com/#applications -> no content in the frame
- https://manage.auth0.com#applications -> no content in the frame

So as how I see it, I don't think they are the same. Maybe I'm wrong? Anyway, just add an extra case to test that scenario where the `#` is after the `/` such as `https://manage.auth0.com/#/applications`.

Issue #1381, opened 2018-05-25
Password Tooltip whitespace on mobile


cocojoe:
As discussed internally.

luisrudge:
![image](https://user-images.githubusercontent.com/941075/40549807-d14a1fd0-600e-11e8-97e0-2f068d8a33ca.png)

PR #1382, opened 2018-05-25
Hide password strength message when the password is valid

luisrudge:
fix https://github.com/auth0/lock/issues/1381

### DIFF
--- a/src/__tests__/field/__snapshots__/password_pane.test.jsx.snap
+++ b/src/__tests__/field/__snapshots__/password_pane.test.jsx.snap
@@ -34,6 +34,7 @@ exports[`PasswordPane disables input when submitting 1`] = `
     data-placeholder="placeholder"
     data-policy="policy"
     data-showPassword="showPassword"
+    data-showPasswordStrenghtMessage={false}
     data-strengthMessages={Object {}}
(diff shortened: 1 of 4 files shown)


src/__tests__/field/__snapshots__/password_pane.test.jsx.snap
```diff
@@ -104,6 +107,27 @@ exports[`PasswordPane sets isValid as true when \`isFieldVisiblyInvalid\` is fal
     data-placeholder="placeholder"
     data-policy="policy"
     data-showPassword="showPassword"
+    data-showPasswordStrenghtMessage={false}
```
lbalmaceda: This property has a typo: `Strenght -> Strength`

src/__tests__/field/password_pane.test.jsx
```diff
@@ -60,6 +61,12 @@ describe('PasswordPane', () => {
 
     expectComponent(<PasswordPane {...defaultProps} />).toMatchSnapshot();
   });
+  it('sets showPasswordStrenghtMessage as true when `isFieldValid` is false', () => {
```
lbalmaceda: what about the opposite case?

PR #1389, opened 2018-05-30
Sending AD information when calling getssodata

luisrudge:
We're not sending the `AD` flag to auth0.js when we call getSSOData. This is fine for embedded since we ignore this flag. But for hosted pages, which use the /ssodata endpoint, this flag is required in order to return information about AD connections.

### DIFF
--- a/src/core/web_api/p2_api.js
+++ b/src/core/web_api/p2_api.js
@@ -123,17 +123,8 @@ class Auth0APIClient {
     this.getUserInfo(token, callback);
   }
 
-  getSSOData(cb) {
-    if (this.isUniversalLogin) {
-      superagent
-        .get(`https://${this.domain}/user/ssodata`)
-        .withCredentials()
-        .end((err, res) => {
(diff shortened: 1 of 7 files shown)


src/core/web_api/p2_api.js
```diff
@@ -123,17 +123,8 @@ class Auth0APIClient {
     this.getUserInfo(token, callback);
   }
 
-  getSSOData(cb) {
-    if (this.isUniversalLogin) {
```
cocojoe: This logic is no longer required?

cocojoe:
@luisrudge please rebase and add some details to the description? 
What was the problem? Why should it be fixed?

src/core/web_api/p2_api.js
luisrudge: nope, auth0.js has all the logic to handle universal login

luisrudge:
@cocojoe done

PR #1390, opened 2018-05-31
Don’t use two-byte character

yuyaohshimo:
Fixed two-byte character in Japanese. Please review/merge 🙇


luisrudge:
Thanks for the PR!

PR #1400, opened 2018-06-07
Fallback to email if the username field is empty

sandrinodimattia:
## Description

Today when you have an enterprise connection configured and you type in an email address that matches the domain of this connection (eg: john@contoso.com => contoso.com), the login/signup/password reset pages will take that into account.

- Login => will change to SSO mode
- Signup => will change to SSO mode
- Password Reset => will prevent you from changing your password with a message saying you need to contact your administrator

While this works fine for database connections, there is an issue when the database connection requires a username. With usernames required, the `databaseUsernameValue` method will only look at the username field. While this works for the login page, it does not for signup and password reset (because there you enter your email address ... in the email field). This currently results in:

- Login => will change to SSO mode
- Signup => will not change to SSO. instead the signup form is still shown.
- Password Reset => will not show a warning.

This change will have `databaseUsernameValue` fallback to the email address if the username field is not filled in. With that, when an enterprise user accidentally fills in their corporate email address on the signup or password reset page, we will trigger the correct behavior of forcing them to use SSO (or preventing them to change their password).

## Testing

The change has manually been tested. An account configured with a DB connection requiring a username + an enterprise connection now results in this behavior:

- Login => will change to SSO mode
- Signup => will change to SSO mode
- Password Reset => will prevent you from changing your password with a message saying you need to contact your administrator

Unit tests were also added.

### DIFF
--- a/src/__tests__/connection/database/index.test.js
+++ b/src/__tests__/connection/database/index.test.js
@@ -1,7 +1,7 @@
 import Immutable, { List, Map } from 'immutable';
 import { databaseUsernameValue } from '../../../connection/database';
 
-describe.only('databaseUsernameValue', () => {
+describe('databaseUsernameValue', () => {
   const getModel = (email, username, usernameRequired) =>
     Immutable.fromJS({
       field: {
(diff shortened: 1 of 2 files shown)


luisrudge:
@sandrinodimattia can I get you to write all my PR descriptions? 😍

Issue #1397, opened 2018-06-07
Lock's SSO transformation triggers LastPass Chrome extension to autofill repeatedly


ethanyoung:
Lock version: 10
Browser: Chrome (tested on Firefox, the issue does not happen on Firefox)
OS: MacOS

When SSO is enabled, Lock will transform to hide the password field when the user-type email matches a domain configured in Auth0. In addition, It will transform to show the password field again when the user starts deleting characters from the email field (domain becomes not matched).

The problem is, if the user installed LassPass Chrome extension, which autofills the email, every time the user updates the email, Lock's transformation triggers LastPass extension to autofill the email field repeatedly. This causes the user cannot update the email field at all, unless the user hit delete furiously.

Here is an example:

![screen shot 2018-06-07 at 12 09 59 pm](https://user-images.githubusercontent.com/6742999/41075034-1d28bf9e-6a4e-11e8-8d24-428c43749fed.png)

LastPass autofills the email, and the domain is matched. The password field is hidden.

![screen shot 2018-06-07 at 12 25 51 pm](https://user-images.githubusercontent.com/6742999/41075036-1f20198c-6a4e-11e8-82f8-40d3083939ff.png)

If the user hit `delete`, Lock should transform to show the password field as shown in the second image. However, the email is autofilled again. Lock immediately becomes back to the first image.

luisrudge:
Can you record a video of this? I don't understand what's happening.

ethanyoung:
A small screen recording can be downloaded [here](https://drive.google.com/open?id=1bRoZI79VGSPlccJ51Oyv5wXDrNkrycIw) (15MB).

luisrudge:
Password managers are weeeeeird :) I just submitted a fix, which should be out in the next release. Thanks for reporting it

ethanyoung:
Thanks!

May I ask when the next release will be? So that we can arrange our release to apply the fix.

luisrudge:
I'm trying to narrow a few extra SSO issues and I to ship everything on Thursday. As always, things might go wrong and dates can change, but those this is the plan right now 😬

luisrudge:
I just opened the last pr for the SSO issues. I'll cut a new release once both are merged! Thanks for your patience.

ethanyoung:
Looks like this fix is not in _11.17.2_. Just want to ask approximately when will 11.17.3 be release, please?

luisrudge:
Sorry to keep you waiting. The next release is scheduled to happen this week.

PR #1407, opened 2018-06-12
In SSO mode, hide the password input instead of removing it from the DOM

luisrudge:
Fix https://github.com/auth0/lock/issues/1397

The issue happened because Lock was REMOVING the password input from the DOM. The fix was to let the input in the DOM, but add a css class that hides the input.

Before:
lastpass

After:
lastpassok

### DIFF
--- a/src/connection/database/login_pane.jsx
+++ b/src/connection/database/login_pane.jsx
@@ -49,10 +49,6 @@ export default class LoginPane extends React.Component {
         />
       );
 
-    const passwordPane = showPassword ? (
-      <PasswordPane i18n={i18n} lock={lock} placeholder={passwordInputPlaceholder} />
-    ) : null;
-
     const dontRememberPassword =
       showForgotPasswordLink && hasScreen(lock, 'forgotPassword') ? (
         <p className="auth0-lock-alternative">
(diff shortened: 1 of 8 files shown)


src/connection/database/login_pane.jsx
```diff
@@ -70,7 +66,12 @@ export default class LoginPane extends React.Component {
       <div>
         {header}
         {fieldPane}
-        {passwordPane}
+        <PasswordPane
```
joshcanhelp: Why are you switching from using a `const`? Not a blocker, just curious

test/helper/ui.js
```diff
@@ -179,6 +179,9 @@ export const hasOneSocialBigButton = hasOneViewFn(
   '.auth0-lock-social-button.auth0-lock-social-big-button'
 );
 export const hasPasswordInput = hasInputFn('password');
```
joshcanhelp: How is this used? Should it account for the input being hidden?

joshcanhelp: Maybe that's handled by the `hasHiddenPasswordInput` you added but still worth a consideration.

src/connection/database/login_pane.jsx
luisrudge: the const was a conditional. It was adding/removing the input from the DOM based on that condition. This triggered the bug itself. So, now, it will always have the input, it will only use css to hide the input

test/helper/ui.js
luisrudge: not sure what you mean 😬

PR #1408, opened 2018-06-12
Print current state of the app when clicking in the header (Only in DEV)

luisrudge:
When you click the header in dev mode, it will show this:
image

This is what gets in the final bundle:
image

Which gets removed when we minify the code:
image


luisrudge:
can't really do this because that would force everyone consuming this package via NPM to inject the global __DEV__  variable.;

PR #1419, opened 2018-06-21
add allowPasswordAutocomplete option

### LINKED ISSUE #1406: allowAutocomplete for password
Setting `allowAutocomplete: true` does not enable autocomplete for the password field, as the name suggests—the property name is not specific, so it should apply to all fields (the [documentation](https://auth0.com/docs/libraries/lock/v11/configuration#allowautocomplete-boolean-) does specify the limitation, which is confusing). Most browsers now ignore this obnoxious attribute, but a few stragglers are behind the times.

[Lock is currently hard-coded](https://github.com/auth0/lock/blob/master/src/ui/input/password_input.jsx#L69) to set `autocomplete="off"`.

Some mobile browsers fail to ignore that attribute, which interferes with password managers.

jshado1:
resolves #1406

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -157,7 +157,9 @@ export function stopRendering(m) {
 function extractUIOptions(id, options) {
   const closable = options.container
     ? false
-    : undefined === options.closable ? true : !!options.closable;
+    : undefined === options.closable
+      ? true
+      : !!options.closable;
   const theme = options.theme || {};
   const { labeledSubmitButton, hideMainScreenTitle, logo, primaryColor, authButtons } = theme;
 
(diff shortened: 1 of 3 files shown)


src/core/index.js
```diff
@@ -157,7 +157,9 @@ export function stopRendering(m) {
 function extractUIOptions(id, options) {
   const closable = options.container
     ? false
-    : undefined === options.closable ? true : !!options.closable;
+    : undefined === options.closable
+      ? true
+      : !!options.closable;
```
jshado1: I did not change this line. I don't know where the change came from.

luisrudge: This is just our prettifier automatically breaking the line

jshado1:
@luisrudge I could not get all of the dependencies to install, so I could not run the app locally or execute the specs.

jshado1:
@luisrudge I don't know what the error in Circle means:

```
Error: At least one database, enterprise or social connection needs to be available.
      at src/core/index.js:489
```

This does not sound related to what I changed.

luisrudge:
Here's the issue: https://github.com/auth0/lock/pull/1419/files#diff-4e772ed67ae3e72cd5b227d85ad75529R51

There's no `lock` in scope at that moment

jshado1:
@luisrudge all the checks are passing. is this good to go?

jshado1:
hiya @luisrudge, do you know when this might get released? I see the last release was about a month ago, alas, just before this made it in.

luisrudge:
I'm aiming for today 🙏

Issue #1434, opened 2018-07-14
Select lose its options when you open the popup second time.


kysonic:
Open this link: 

https://codepen.io/anon/pen/RBWmzZ

Steps to reproduce: 

1. Go to sign up tab. Click on "test" select. Everything is ok.
2. Close auth0 popup.
3. After it opens, go to sign up tab again 
4. Click on "test" again. It doesn't work 

http://joxi.ru/KAxddwzuM3LMbA

PR #1448, opened 2018-07-25
Fix crash when showing lock for the second time with custom select input

luisrudge:
fix https://github.com/auth0/lock/issues/1434


luisrudge:
@joshcanhelp no need, really. I could reproduce their use case locally.

PR #1471, opened 2018-08-22
Improve accessibility

andrew-me:
Includes:

1) Adds lang attribute to container div. Ideally we'd set lang on the html tag. However lock can appear as a modal in an external sites html, so next best thing is to set lang on the container div
2) Adjusts the markup and styling of the 'Login' / 'Signup' tabs to improve keyboard navigation. The tabs now have role 'navigation. If a tab is focused it has different styling. The current tab is no longer a link
3) Some svgs have been give aria-hidden"true". They are decoration so hiding them from screen readers seems sensible
4) The box Button component has been changed from a 'span' to a 'button' so that it can be navigated to with a keyboard. Also, it's been styled to look different when focused.
5) The submit button has been given a label
6) Empty alt tag on the logo img in the header (it's arguably decoration)
7) Aria tags on form inputs
8) Linking of form errors to their inputs via aria. Use of role='alert' on form errors for screenreaders.


luisrudge:
Damn! Thanks! I'll try to review your PR this week, although I'm a bit full of work right now 😬 So, next week at the latest, I'll get this merged. Again, thank you so much!

andrew-me:
Thanks! There are definitely something things that could be added/improved (for example it's not obvious enough which items have keyboard focus) but hopefully it's a good start.

luisrudge:
Hi @andrew-me. Is there anything in this PR that you think is causing this issue? https://github.com/auth0/lock/issues/1477

PR #1472, opened 2018-08-25
Support new minimum password length parameter

luisrudge:


package.json
```diff
@@ -88,6 +88,7 @@
   },
   "dependencies": {
     "auth0-js": "^9.7.3",
+    "auth0-password-policies": "auth0/auth0-password-policies#v1.0.1",
```
lbalmaceda: I'd trust more on password-sheriff than this repo. Sheriff already defines them here: https://github.com/auth0/password-sheriff/blob/4a17c9cba614ed8b75affc7c53ff0d4ba31a5b30/index.js. If this is not possible, then please copy that file and keep it here on this repo.

src/__tests__/core/client/index.test.js
```diff
@@ -0,0 +1,50 @@
...
+                passwordPolicy: 'low'
+              }
+            ]
+          }
+        ]
+      };
+      const result = initClient(Immutable.fromJS({}), client).toJS();
+      expect(result.client.connections.database[0].passwordPolicy).toMatchObject({
+        length: {
+          minLength: 6
```
lbalmaceda: Please add the same tests for the remaining policies.

src/field/password.js
```diff
@@ -1,8 +1,11 @@
-import createPolicy from 'password-sheriff';
+import PasswordPolicy from 'password-sheriff/lib/policy';
 import { setField } from './index';
 
 export function validatePassword(password, policy) {
-  return createPolicy(policy).check(password);
+  if (!policy) {
+    return true;
+  }
+  return new PasswordPolicy(policy.toJS()).check(password);
```
lbalmaceda: Shouldn't this be safer, e.g. checking that if `policy == undefined` then `none` policy is still applied? https://github.com/auth0/password-sheriff/blob/master/index.js#L57 Now it will work since `none` is just `length > 0` but if that changes this code will need to be changed as well. Let's reference this to the `none` policy instead.

src/ui/input/password/password_strength.jsx
```diff
@@ -1,12 +1,12 @@
...
 import React from 'react';
-import createPolicy from 'password-sheriff';
+import PasswordPolicy from 'password-sheriff/lib/policy';
 import util from 'util';
 
 export default class PasswordStrength extends React.Component {
   render() {
     const { password, policy, messages } = this.props;
-    const analysis = createPolicy(policy).missing(password);
+    const analysis = new PasswordPolicy(policy.toJS()).missing(password);
```
lbalmaceda: I think the same applies here

package.json
luisrudge: our change password widget also uses this repo. This is fine

src/field/password.js
luisrudge: policy is only undefined when it's a login, which we don't validate passwords.

src/ui/input/password/password_strength.jsx
luisrudge: this only runs on signup, so we're fine

src/__tests__/core/client/index.test.js
luisrudge: That would be just a test for an external resource, not sure that's needed.

lbalmaceda: How can I trust that your code will behave the same when the policy used is other than `low`? Please add them.

luisrudge: pushed

PR #1478, opened 2018-09-08
Use span for icon buttons

andrew-me:
Fixes regression in v11.9.0

In 11.9.0, for accessibility, some 'span' tags were switched to 'button'.
However hitting the enter key triggers the onClick on buttons. (not sure where this is in the code, but it might be worth changing)
So switching back to 'span' for now.


luisrudge:
Thanks @andrew-me!

PR #1485, opened 2018-09-18
Add `showTerms` option

luisrudge:
Add `showTerms` option to show/hide terms string in the signup page


cocojoe:
Add description, why are we making this change? Should add screenshots as it will change the default display behaviour with the additional box.

luisrudge:
@cocojoe updated

cocojoe:
@luisrudge what happens if `showterms` is `false` but `mustAcceptTerms` is `true`?

luisrudge:
Doesn’t show the terms Because if you have to manually set showTerms to false

Luís
________________________________

cocojoe:
OK that seems fair as the user has explicitly disabled them.

Issue #1484, opened 2018-09-18
checkSession drops scopes from original lock constructor


ediphy-azorab:
Contrary to the docs, calling `lock.checkSession({}, cb)` does not get back a token with the same scopes as when initially connecting. 

Calling `lock.checkSession({ scope: myScopes }, cb)` does work. Since the constructor of the lock takes the scopes in `auth.params.scope`, I suspect that the issue is caused by this difference in convention.

This was observed in lock 11.9.1

PR #1486, opened 2018-09-18
Fix different scope in checkSession

luisrudge:
fix https://github.com/auth0/lock/issues/1484

### DIFF
--- a/src/__tests__/core/web_api/p2_api.test.js
+++ b/src/__tests__/core/web_api/p2_api.test.js
@@ -126,6 +126,14 @@ describe('Auth0APIClient', () => {
         });
         expect(client.authOpt.nonce).toBe('foo');
       });
+      it('should set scope from options.params.scope', () => {
+        const client = getClient({
+          params: {
+            scope: 'foo'
+          }
+        });
+        expect(client.authOpt.scope).toBe('foo');
+      });
     });
   });
   describe('logIn', () => {
--- a/src/core/web_api/p2_api.js
+++ b/src/core/web_api/p2_api.js
@@ -30,6 +30,8 @@ class Auth0APIClient {
       nonce = opts.params.nonce;
     }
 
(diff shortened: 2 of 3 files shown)


src/core/web_api/p2_api.js
```diff
@@ -43,14 +45,16 @@ class Auth0APIClient {
       _sendTelemetry: opts._sendTelemetry === false ? false : true,
       _telemetryInfo: opts._telemetryInfo || default_telemetry,
       state,
-      nonce
+      nonce,
+      scope
     });
 
     this.authOpt = {
```
cocojoe: How does this feed into `checkSession` function?

luisrudge: everything that goes in the constructor, goes to all the calls

cocojoe: checkSession simply calls `logIn` then?

luisrudge: no, sorry. they're separate calls. I just fixed logIn as well, since it would have the same issue

adding to the constructor: fixes the reported issue

adding to authOpt: fixes the same issue, but with a login call

src/__tests__/core/web_api/p2_api.test.js
```diff
@@ -126,6 +126,14 @@ describe('Auth0APIClient', () => {
         });
         expect(client.authOpt.nonce).toBe('foo');
       });
+      it('should set scope from options.params.scope', () => {
+        const client = getClient({
+          params: {
+            scope: 'foo'
+          }
+        });
+        expect(client.authOpt.scope).toBe('foo');
```
cocojoe: This only tests the param was set, can we be sure it made its' way to the authorize URL correctly?

luisrudge: added missing test value

PR #1487, opened 2018-09-19
Added promotion flag and serialization of fields marked with is_promo…

solidmartin:
Added promotion flag and serialization of fields marked with is_promotion into the promotion field. Requires an unpacking rule in Auth0.

It works like this: When setting additional sign-up fields in the lock widgets configuration,
a field can be flagged with is_promotion: true and before the request to signup i sent, those fields are packed, serialized and put in a promotion field and not available directly as a field. 

The lock fields can later be unpacked and processed using a Auth0 Rule

### DIFF
--- a/src/connection/database/actions.js
+++ b/src/connection/database/actions.js
@@ -63,11 +63,16 @@ export function signUp(id) {
 
     if (!additionalSignUpFields(m).isEmpty()) {
       params.user_metadata = {};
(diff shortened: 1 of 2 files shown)


luisrudge:
Hi, thanks for the PR, but this is something specific to your app and it doesn't make sense to add to lock itself.

PR #1488, opened 2018-09-26
Added Email confirmation input 

worgarside:
This PR is to allow users to repeat their email address when signing up as a way of confirmation/validation. We found that a lot of our users are prone to mis-typing their email addresses and this can take a lot of work to resolve. Our fix for this was to add an `additionalSignUpField` with a custom validator but this was a bit of a hack and needed funky css to change the ordering, so we thought we'd do it the right way. :) 

While the functionality is there in a basic form, we're aware that we need to add some tests and it might need a bit of a tidy. Your thoughts would be very much appreciated. Thanks!

### DIFF
--- a/src/connection/database/actions.js
+++ b/src/connection/database/actions.js
@@ -47,6 +47,7 @@ export function signUp(id) {
   const m = read(getEntity, 'lock', id);
   const fields = ['email', 'password'];
   if (databaseConnectionRequiresUsername(m)) fields.push('username');
+  if (l.confirmEmailInput(m)) fields.push('confirmEmail');
   additionalSignUpFields(m).forEach(x => fields.push(x.get('name')));
 
   validateAndSubmit(id, fields, m => {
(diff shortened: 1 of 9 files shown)


luisrudge:
@worgarside Hi 👋 Thanks for the PR!

I just want to give you some feedback on this. I just sent this to our product team to ask them if if this is something we're interested in. I'll get back to you when I hear back from them.

LewisRogers:
Hi 👋 @luisrudge - I work with @worgarside and wondering of you had any update from the above? We would also like to point out that this functionality is purely optional and is passed to the Lock constructor as part of the options object. 

As @worgarside said above, this has been a particular sore spot for some of our customers and has had quite an impact on our operational team which need to then go in and switch emails after a user tries to sign up multiple times. 

The additionalSignUp field option, whilst it does work, feels 'hacky' and breaks the tabindex of the component if we want to render the fields in a logical order. There is also the possibility that future updates will completely ignore these additional css styles that have been used to hack the render order in our current implementation. 

Looking forward to your response and we are both obviously more than happy to take your steer on what we need to do in terms of test / tidy up to make this happen. 

Loving the work and the product by the way. 

Thanks, 

Lewis.

PR #1492, opened 2018-10-01
Improve labelling

Splact:
- Let the user define labels for additional signup fields
- Set Alt attribute on icon 
- When we use `name` as `aria-label` now we give higher priority to the new `label` property


Splact:
All done :)!

luisrudge:
Thanks for the PR! 🎉

PR #1496, opened 2018-10-03
Add default message for Sign-Up Terms & Privacity Policity

### LINKED ISSUE #1482: Sign-Up Terms & Conditions are empty by default
Lock widget is missing a default value for the sign-up _Terms & Conditions_ which causes the container to be hidden. **I think it's important to be straightforward on what the user is signing up to.**

[This line](https://github.com/auth0/lock/blob/7178d6223f02a990889cd44b957abdb6b51b29d5/src/i18n/en.js#L106) should have a text like:

```
By signing up, you agree to our terms of service and privacy policy
```

In the mobile SDKs we also link to the policies (which are customizable, see below). By default, we could display plain text or link to the Auth0 ones: https://auth0.com/privacy and https://auth0.com/terms.

**iOS:** https://github.com/auth0/Lock.swift/blob/master/Lock/Base.lproj/Lock.strings
**Android:** https://github.com/auth0/Lock.Android/blob/master/lib/src/main/res/values/strings.xml

image

rafaelalmeidatk:
Fixes #1482. I am trying to generate the translations but the command is failing with `UnhandledPromiseRejectionWarning: Error: Service Unavailable`, it didn't translated for `ro` and bellow, I will try again later.

On a side note: I have been looking for ways to add customizable address on the message (just like Android) but this would require creating a default config file (for fallbacks) and I am not sure where this should go (maybe `./src/defaultOptions.js`?)

### DIFF
--- a/src/i18n/en.js
+++ b/src/i18n/en.js
@@ -103,7 +103,8 @@ export default {
   signUpTitle: 'Sign Up',
   signUpLabel: 'Sign Up',
   signUpSubmitLabel: 'Sign Up',
-  signUpTerms: '',
+  signUpTerms:
+    'By signing up, you agree to our <a href="https://auth0.com/terms" target="_new">terms of service</a> and <a href="https://auth0.com/privacy" target="_new">privacy policy</a>',
   signUpWithLabel: 'Sign up with %s',
   socialLoginInstructions: '',
   socialSignUpInstructions: '',


luisrudge:
We already have a PR open for the strings: https://github.com/auth0/lock/pull/1485


> On a side note: I have been looking for ways to add customizable address on the message (just like Android) but this would require creating a default config file (for fallbacks) and I am not sure where this should go (maybe ./src/defaultOptions.js?)

I'm not sure what address you mean..

rafaelalmeidatk:
@luisrudge sorry for that I didn't see on the issue. About the address: on Android you can change the address of the links on the message (https://auth0.com/privacy etc) with a configuration but this is not so necessary because the link won't change so often (may not even change at all).

PR #1504, opened 2018-10-11
[Sustainment] Fixed Reset Password Avatar/Enterprise Issue

cocojoe:
**Description**
When `avatar: null` is specified in the the Lock configuration options, reset password will not display the enterprise error when an enterprise domain is detected. 

avatar_null

**References**
ZD #48775

**Notes**
This is a hot patch, I feel a more complete solution that addresses the root cause that is rapid duplicate calls to the [isSubmitDisabled](https://github.com/auth0/lock/blob/fixed-enterprise-null-avatar/src/connection/database/reset_password.jsx#L43) method should be looked at. Although this requires more in depth platform knowledge.


luisrudge:
Sorry. Yesterday github was having all sorts of issues with CIs, webhooks etc. It'll be released today.

PR #1505, opened 2018-10-11
Add passwordless support to prefill

luisrudge:
The `prefill` option makes the input text to be automatically filled with the given value. For example

#### Will initialize the email input with `foo@bar.com`
```js
prefill: {
  email: 'foo@bar.com'
}
```

#### Will initialize the phone input with `489999999`
```js
prefill: {
  phoneNumber: '489999999'
}
```

This was already supported by Auth0Lock. This PR adds support for the same behavior in Auth0LockPasswordless.


src/engine/passwordless.js
```diff
@@ -80,15 +94,6 @@ class Passwordless {
     } else {
       return passwordlessStarted(m) ? new VcodeScreen() : new SocialOrPhoneNumberLoginScreen();
     }
-
-    setTimeout(() => {
```
luisrudge: this was unreachable code

src/engine/passwordless.js
```diff
@@ -16,6 +16,19 @@ import { isDone, isSuccess } from '../sync';
...
+import { setEmail } from '../field/email';
+import { setPhoneNumber } from '../field/phone_number';
+
+const setPrefill = m => {
+  const { email, phoneNumber } = l.prefill(m).toJS();
+  if (typeof email === 'string') {
+    m = setEmail(m, email);
+  }
+  if (typeof phoneNumber === 'string') {
+    m = setPhoneNumber(m, phoneNumber);
```
lbalmaceda: the previous `m` value get's overwritten with this one if both if clauses are called. Is this intended? what means the `m` returned?

src/__tests__/engine/passwordless.test.js
```diff
@@ -0,0 +1,52 @@
...
+        setEmail: jest.fn(m => m)
+      }));
+      jest.mock('field/phone_number', () => ({
+        setPhoneNumber: jest.fn(m => m)
+      }));
+    });
+    it('when prefill options has `email` value', () => {
+      const engine = getEngine();
+      engine.didReceiveClientSettings('model');
+      expect(require('field/email').setEmail.mock.calls.length).toBe(1);
```
lbalmaceda: would it make sense to take this `require(..)` to the top of the file, so to resolve it only once? (even if it's a test, every second counts toward ci running limits)/

Same for the require below

src/engine/passwordless.js
luisrudge: yeah, that's how lock's internal state work. `m` is an immutable object, which means that each `setSOMETHING` method returns a new instance of the model

src/__tests__/engine/passwordless.test.js
luisrudge: pushed

PR #1513, opened 2018-10-19
added a link to intro video justification in the extended description

saicodes:
I felt that a link to your into video on your GitHub repo readme page would help out some beginners who stumble upon your repo to understand what the auth0 really does.
added a link in the form of key emoji(to the readme) to the into video so as to keep with the pace of today's world of emoji communication and the key was chosen because for obvious reason(You work on authentication).
feel free to change out the font or wordings and spacing but I think a link to intro video is very useful


luisrudge:
Thanks, but I think linking to auth0.com is enough.

PR #1514, opened 2018-10-19
Handle HTML in the `invalidHint` better

glidej:
The "hrd.not_matching_email" `invalidHint` is coming into `InputWrap` as HTML, presumably as intended, but the `InputWrap` component is not expecting that. This causes a couple things:

- Locally, it throws an exception, because the PropType of `string` is violated occasionally.
- Because the component is otherwise unaware that HTML might be coming in, the css was written with a generic selector (`span`) to apply box/border styling. This causes double padding if HTML comes in pre-packaged as a span.

This PR gives that component some flexibility to deal with the possibility that fully formed html might be coming in.

It looks like most, if not all, of the `<input>_pane` components could be updated to always pass html to be consistent. `CustomInput` would need some love, too. 💭 

(Before) Double padding around the Hint:
screen shot 2018-10-19 at 11 31 42 am

(Before) Exception:
screen shot 2018-10-19 at 11 31 47 am

(After) Consistent padding:
screen shot 2018-10-19 at 11 27 59 am

screen shot 2018-10-19 at 11 27 52 am

### DIFF
--- a/src/ui/input/input_wrap.jsx
+++ b/src/ui/input/input_wrap.jsx
@@ -31,7 +31,7 @@ export default class InputWrap extends React.Component {
     const errorTooltip =
       !isValid && invalidHint ? (
         <div role="alert" id={`auth0-lock-error-msg-${name}`} className="auth0-lock-error-msg">
-          <span>{invalidHint}</span>
+          {invalidHint}
         </div>
       ) : null;
 
@@ -55,7 +55,7 @@ InputWrap.propTypes = {
     PropTypes.arrayOf(PropTypes.element).isRequired
   ]),
   focused: PropTypes.bool,
-  invalidHint: PropTypes.string,
+  invalidHint: PropTypes.node,
   isValid: PropTypes.bool.isRequired,
   name: PropTypes.string.isRequired,
   svg: PropTypes.string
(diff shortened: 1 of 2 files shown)


src/ui/input/input_wrap.jsx
```diff
@@ -55,7 +55,7 @@ InputWrap.propTypes = {
     PropTypes.arrayOf(PropTypes.element).isRequired
   ]),
   focused: PropTypes.bool,
-  invalidHint: PropTypes.string,
+  invalidHint: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
```
luisrudge: you can use just `PropTypes.node` here, because it already includes strings

glidej: oh, that's great, thanks for the tip. I'll update.

luisrudge:
```suggestion
  invalidHint: PropTypes.node,
```

luisrudge:
Thanks for the PR! 🎉

glidej:
@luisrudge thanks for the quick approval/merge. do you have a cadence for releases?

luisrudge:
should have happened yesterday but got delayed :( I'll release on Monday

glidej:
@luisrudge sorry to badger, I'm looking to schedule our own production release and I'd like to include this change. do you have an updated timeline for your next release?

luisrudge:
Sorry. Yesterday github was having all sorts of issues with CIs, webhooks etc. It'll be released today.

PR #1515, opened 2018-10-19
Minor fixes and changed var to const

ashishpatelcs:


luisrudge:
Thanks, but var is fine considering not every browser supports const (we still support IE10 😢)

ashishpatelcs:
Oh, thank you!

PR #1516, opened 2018-10-21
improved German translation

ritterasdf:
improved German translation

a few small improvements to bring the German translation closer to the English original and to increase consistency, such as:
1. email address is now E-Mail-Adresse, which is different from email message (E-Mail).
2. Multifactor is now Multifaktor-Authentifizierung everywhere
3. "Es tut uns leid" is placed in all messages, that have a "we're sorry" in English.
4. sentLabel is now correct.
5. Corporate is now "geschäftlich" everywhere instead of a mix of "Unternehmen" and "geschäftlich".
6. changed lines to reflect same ordering as English

added a minor comma to the English original

### DIFF
--- a/src/i18n/de.js
+++ b/src/i18n/de.js
@@ -5,7 +5,7 @@ export default {
         'Sie haben das Limit für Rücksetzungsversuche des Passworts erreicht. Bitte warten Sie, bevor Sie es erneut versuchen.',
       'lock.fallback': 'Es tut uns leid, beim Zurücksetzen des Passworts ist ein Fehler aufgetreten.',
       enterprise_email:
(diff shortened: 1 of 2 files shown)


src/i18n/de.js
```diff
@@ -101,7 +102,9 @@ export default {
   resendLabel: 'Erneut senden',
   resendingLabel: 'Wird erneut gesendet...',
   retryLabel: 'Wiederholen',
-  sentLabel: 'Senden',
+  sentLabel: 'Gesendet!',
+  showPassword: 'Passwort anzeigen'
```
luisrudge: you're missing a `,` at the end of the line:

```suggestion
  showPassword: 'Passwort anzeigen',
```

ritterasdf: thank you for catching that. This used to be the last line of the document, which didn't end in a comma but I moved it up to have the same structure as the English version.

luisrudge:
I'll check this review with a german co-worker and get back to you. Thanks!

Issue #1174, opened 2017-11-21
add ids to inputs and buttons so it's easier to automate tests


luisrudge:

luisrudge:
@verde79 absolutely! 😬 This was marked as low priority to us, but it's definitely something we need 🎉

verde79:
@luisrudge for clarification, is it just the playground/index.html that needs id's or are there others I'm not seeing?  Also is there any specific naming conventions for your id's?

luisrudge:
@verde79 no, it's actually the IDs inside lock itself.

For example: 
- https://github.com/auth0/lock/blob/ab7c94ff0f813227d667b2059028282b89dc926f/src/ui/input/text_input.jsx
- https://github.com/auth0/lock/blob/ab7c94ff0f813227d667b2059028282b89dc926f/src/ui/input/checkbox_input.jsx
- https://github.com/auth0/lock/blob/1c9bcc22b861abb9676b8a2247f3f117a10fd7d7/src/ui/input/mfa_code_input.jsx

luisrudge:
fixed in #1517

PR #1517, opened 2018-10-22
Add ids to inputs and buttons

tingaloo:
Took a stab at #1174, added ids to inputs. I assumed that there will be one instance of some inputs, correct me if i'm wrong.

### DIFF
--- a/src/ui/box/button.jsx
+++ b/src/ui/box/button.jsx
@@ -10,6 +10,7 @@ const svgs = {
 
 const IconButton = ({ name, onClick, svg }) => (
   <span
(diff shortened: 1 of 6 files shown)


tingaloo:
in ```chrome.jsx``` there is a ```<Header/>``` that contains ```<BackButton />``` However header does not have model or lock defined so i cannot pass that to BackButton, whereas other components do so.

luisrudge:
@tingaloo you're just missing the snapshots! Almost there 😅 

To fix snapshots, run:

```
yarn test:jest -- -u
```

Issue #1537, opened 2018-10-31
Lock 11.4 prevents sign in for custom db usernames including a space


mmaddex:
Starting with 11.4 typing a space character in the username input box was disabled. However it's possible to have a Custom DB with usernames containing spaces. Newer versions of Lock should support this use case.

luisrudge:
ref: https://github.com/auth0/lock/pull/1280

Are you saying that you can't **end** a string with a space or you can't even type a space in between two words?

mmaddex:
Sorry I should have included more detailed repro steps. I am saying that I cannot type a space between two words.

PR #1541, opened 2018-11-06
Trim username only when it looks like an email

luisrudge:
fix https://github.com/auth0/lock/issues/1537

### Changes
Instead of always trimming the username, we'll only trim when the username looks like an email.

### References

https://github.com/auth0/lock/issues/1537

### Testing


luisrudge:
We'll do this in another way

cocojoe:
@luisrudge can you reference the alternative PR please, helps anyone originally following. Thanks

luisrudge:
https://github.com/auth0/lock/pull/1541

cocojoe:
@luisrudge you just referenced the same PR...

luisrudge:
😝 sorry. https://github.com/auth0/lock/pull/1546

PR #1546, opened 2018-11-12
Trim auth params before sending to the API (not while typing)

luisrudge:
fix https://github.com/auth0/lock/issues/1537

### Changes

Instead of trimming email/username/phoneNumber while the user is typing, we're now trimming them when we send them to the server.

### References

https://github.com/auth0/lock/issues/1537

### Testing


cocojoe:
@luisrudge can you clarify your description please, you trim on submission of the field or it's only trimmed when an API request is made?

luisrudge:
@cocojoe I only trim the value being sent to the API. The input value is left untouched.

src/core/web_api/helper.js
```diff
@@ -134,3 +134,13 @@ export function webAuthOverrides({ __tenant, __token_issuer, __jwks_uri } = {})
   }
   return null;
 }
+
+export function trimAuthParams(params = {}) {
+  const { ...p } = params;
+  ['username', 'email', 'phoneNumber'].forEach(k => {
```
lbalmaceda: what about MFA code?

src/core/web_api/p2_api.js
```diff
@@ -65,11 +68,13 @@ class Auth0APIClient {
     // TODO: for passwordless only, try to clean in auth0.js
     // client._shouldRedirect = redirect || responseType === "code" || !!redirectUrl;
     const f = loginCallback(false, this.domain, cb);
-    const loginOptions = normalizeAuthParams({
-      ...options,
-      ...this.authOpt,
-      ...authParams
-    });
+    const loginOptions = trimAuthParams(
```
joshcanhelp: The function wrap feels ... wrong. Can `trimAuthParams` be a part of `normalizeAuthParams`? What if we need to do another transform here? Maybe there should be a `prepareParams` or similar to handle everything?

src/core/web_api/p2_api.js
```diff
@@ -100,20 +105,16 @@ class Auth0APIClient {
...
-
     delete options.autoLogin;
-
-    this.client.signup(options, (err, result) => cb(err, result));
+    this.client.signup(trimAuthParams(options), (err, result) => cb(err, result));
   }
 
   resetPassword(options, cb) {
-    this.client.changePassword(options, cb);
+    this.client.changePassword(trimAuthParams(options), cb);
```
joshcanhelp: Similar feedback here ... there might be another transform in the future, in which case this starts to get really hard to read.

PR #1539, opened 2018-11-06
Fix mobile styles using different heights

luisrudge:
fix #1474 

### Changes

Before: 
image

After:
image

### References

#1474 
#824
#1223

### Testing

* [X] Visual testing

PR #1547, opened 2018-11-12
Fix padding in all screens

luisrudge:
### Changes

When we did https://github.com/auth0/lock/pull/1539, I had to create another div to wrap the internal content. It was copy/paster in every "screen" component, but I forgot in some of them. Instead of continue the copy/paste, I added the padding to the screen's wrapper component. This will have the same effect and it works with all screens (current and new) OOTB.

### References

https://github.com/auth0/lock/pull/1539

### Testing

### DIFF
--- a/css/index.styl
+++ b/css/index.styl
@@ -308,6 +308,7 @@ loadingSize = 30px
     -webkit-box-sizing border-box
     -moz-box-sizing border-box
     box-sizing border-box
+    padding 20px
 
   .auth0-lock-form
     display initial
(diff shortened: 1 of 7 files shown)


luisrudge:
in hindsight, the copy/paste was the wrong approach anyway 😬 The padding is now [being added](https://github.com/auth0/lock/pull/1547/files#diff-6e75cf9db4a3c7d943f8603dc182c0e9R307) to the div that wraps all screens, so we're good now.

Issue #1556, opened 2018-11-23
Lock v11.12.0 throws an error on login


olegk101:
After upgrading from 11.11 to 11.12 I started to get the following error on login

````
index.js:46 Uncaught TypeError: t.get is not a function
    at Object.o [as id] (index.js:46)
    at c (confirmation_pane.jsx:8)
    at p._constructComponentWithoutOwner (ReactCompositeComponent.js:306)
    at p._constructComponent (ReactCompositeComponent.js:282)
    at p.mountComponent (ReactCompositeComponent.js:185)
    at Object.mountComponent (ReactReconciler.js:43)
    at performInitialMount (ReactCompositeComponent.js:368)
    at p.mountComponent (ReactCompositeComponent.js:255)
    at Object.mountComponent (ReactReconciler.js:43)
    at performInitialMount (ReactCompositeComponent.js:368)```

````

```
export function id(m) {
  return m.get('id'); <-- fails here
}
```

Used the following params

```
<head>
    <script src="https://cdn.auth0.com/js/lock/11.12.0/lock.min.js"></script>
</head>
<body>
    <script>
        const lock = new Auth0Lock(
        'CLIENT_ID',
        'AUTH0_DOMAIN',
        {
            "auth": {
                "redirect":false,
                "audience":"AUDIENCE"
            },
            "rememberLastLogin":false,
            "allowSignUp":false
        });

        lock.show()
</script>
</body>
```

adding this piece of code fixed it

```
window.lock = lock;
window.lock.get = () => ''
```

luisrudge:
Fix released in 11.12.1

PR #1557, opened 2018-11-23
Fix `lock is undefined` error in non-redirect scenarios

luisrudge:
fix https://github.com/auth0/lock/issues/1556

### Changes

Passing the lock instance as props so the confirmation pane (only shown when redirect:false) doesn't throw an error.

### References

https://github.com/auth0/lock/issues/1556


cocojoe:
How is this tested?

PR #1590, opened 2019-01-21
Added an i18n file for the Greek language Ελληνικά.

anthony0030:
I added an i18n file for the Greek language Ελληνικά. It is useful for the company that I work at
because we are located in Greece and our clients speak Greek.

Issue:
[https://github.com/GoThassos/gothassos-rails/issues/135](https://github.com/GoThassos/gothassos-rails/issues/135) (private)


luisrudge:
Ευχαριστώ 🎉

PR #1597, opened 2019-01-29
Add visual regression testing using Applitools Eyes

myspivey:
### Changes

This PR adds support for Visual Regression testing using Applitools Eyes. This is integrated into CircleCI to better support the PR process and development of Lock. If you are unfamiliar with Visual Regression testing, I have included in the references a link to a high level overview from Applitools. As an open source tool, this service will be offered to you free with an OSS license. They do offer a generous free tier if you want to get a license now to play around with it. 

I picked Cypress as the browser test harness which Applitools already has an SDK for. This allowed me to quickly and succinctly create a visual test suite that is easily maintainable. Through the nature of how it grabs each example this should automatically expand to any future iterations you make.

The process for this testing is the docs site is served using concurrently and the existing design script, then Cypress launches a headless Chrome where it runs the tests. Applitool's SDK then takes DOM snapshots w/ styles (not screenshots) from that headless Chrome and sends them to Applitool's servers where it renders those results in Chrome and Firefox and then takes a screenshot of that result and does its intelligent comparisons against the baseline. The baseline is created the first time you run it. If there's an issue, it'll be reported in CircleCI with an included link directly to Applitool's web app where you can see the visual differences and either accept or reject them or do other cool things like set ignored regions or tweak the matching behavior.

If you have any other questions I am happy to help! Thanks for your time!

### References

https://applitools.com/docs/topics/overview/overview-visual-testing.html
https://applitools.com/blog/troubleshoot-fix-react-bugs
https://circleci.com/gh/myspivey/lock/6

### Testing

This is a new test suite that uses the /support/design folder to run a local server, generate the DOM and then send this content to Applitools Eyes to render and visually compare. This will fail the CircleCI build as at this time there is not an environment api key provided. I have tested this on my own instance of CircleCI which you can find here, https://circleci.com/gh/myspivey/lock/6.

### Checklist

* [ X ] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
* [ X ] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)
* [ X ] All existing and new tests complete without errors
* [ X ] All code quality tools/guidelines have been run/followed
* [ X ] All relevant assets have been compiled
* [ X ] All active GitHub checks have passed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,9 @@ jobs:
       - run: 
           name: Tests
           command: yarn run test:jest
+      - run:
+          name: Visual Tests
+          command: yarn run test:visual
       - store_artifacts:
           path: build
 
--- a/circle.yml
+++ b/circle.yml
@@ -14,6 +14,7 @@ test:
   override:
     - yarn run test:cli
     - yarn run test:jest
+    - yarn run test:visual
 general:
   artifacts:
     - build
--- a/cypress.json
+++ b/cypress.json
@@ -1,0 +1,4 @@
+{
+  "eyesTimeout": 1200000,
+  "video": false
+}
--- a/cypress/.eslintrc.json
+++ b/cypress/.eslintrc.json
@@ -1,0 +1,11 @@
+{
+  "env": {
+    "browser": true,
+    "node": true,
+    "mocha": true
+  },
+  "globals": {
+    "cy": true
+  },
+  "plugins": ["mocha"]
+}
--- a/cypress/integration/visual-regression-smoke-tests.spec.js
+++ b/cypress/integration/visual-regression-smoke-tests.spec.js
@@ -1,0 +1,43 @@
+describe('Visual Regression - Smoke Tests', () => {
+  beforeEach(() => {
+    cy.eyesOpen({
+      appName: 'Auth0 Lock',
+      testName: 'Visual Regression - Smoke Tests',
+      browser: [
+        {
+          name: 'chrome',
+          width: 1280,
+          height: 800
+        },
+        {
+          name: 'firefox',
+          width: 1280,
+          height: 800
+        }
+      ]
+    });
+  });
+
+  afterEach(() => {
+    cy.eyesClose();
+  });
+
+  it('Should render each lock style correctly', () => {
+    cy.visit(`http://localhost:8080/support/design/`);
+    //Delay DOM grab till locks render
+    cy.wait(2000);
+
+    cy.get('.lock-container').then($nodes => {
+      for (let i = 0, l = $nodes.length; i < l; i++) {
+        cy.eyesCheckWindow({
+          sizeMode: 'selector',
+          selector: {
+            type: 'xpath',
+            selector: `//*[@id='${$nodes.eq(i).get(0).id}'][1]`
+          },
+          tag: $nodes.eq(i).get(0).id
+        });
+      }
+    });
+  });
+});
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -1,0 +1,19 @@
+// ***********************************************************
+// This example plugins/index.js can be used to load plugins
+//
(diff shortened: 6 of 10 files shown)


cocojoe:
@myspivey Thanks for the contribution we will take a look into this, in the mean time can you rebase your branch please with latest and ensure CI passes.

luisrudge:
Thanks @myspivey. At Auth0, we're already using https://www.chromaticqa.com/ in other projects (https://github.com/auth0/cosmos). Do you have any experience with it? Can you do a comparison between them?

myspivey:
Hey LuÍs,
I do not have any first-hand experience with Chromatica. I spent a bit of time checking out what they have and very cool stuff indeed! I would say from my understanding of Applitools Eyes and Chromatica, there are a couple of unique differences that set them apart.

The first and probably most important is that they currently only support Storybook for visual testing. Storybook is an amazing tool and does indeed make visual testing much easier. Applitools also has a Storybook API that functions in much the same way. For this PR and repo however as it does not use Storybook, Chromatica would not be able to work. This PR makes use of the Applitools Cypress API which was trivial to implement as opposed to Storybook which would be a more substantial amount of effort and complexity to implement. This as you will note in the PR limits the amount of changes and things that the Auth0 and OSS community would have to maintain for this visual testing implementation.

The second thing that stands out relates to the first. With Storybook you are limited to what Storybook presents with no real ability from the testing suite to modify that style programmatically. With Cypress and Applitools eyes, you could add additional tests that click or otherwise interact with Lock, and then record a DOM snapshot to verify that it visually looks as it should. I noticed you used Zuul to do this from a non-visual perspective already so could be a great additional test to further supplement. Basically, anything you can do with Cypress from an E2E perspective you can also visually test. I did not add this on the PR as I was unsure if that would be indeed wanted, but it is very possible with the tooling in this PR to do.

The third thing I believe to be an advantage is the way in which the visual regression comparisons are done. While they both send a DOM snapshot to be rendered, Applitools uses a complex intelligent algorithm to avoid false positives. 1 to 1 pixel compares can and often are very brittle. Applitools algorithms reduce and eliminate false positives and negatives saving the team time or the dreaded “approve all”. They have a root cause analysis tool that once these items are eliminated can help you tell exactly what was changed to cause the test to fail which is immensely helpful and a huge time saver in correcting the problem. I have included below some blog posts from the Applitools blog that go a bit more in depth into how this comparison works and why it can save time as well as some of the features such as the ability to select how much intelligence to apply to the comparison.

Lastly, and this seems to be where they are getting towards parity, currently Chromatica only supports Chrome while Applitools supports far more browsers. It does appear from their website however that they are working to support more as well so not really a long-term advantage. I would also note they talk about on their site focusing on components for testing, you will note in this PR we are also doing exactly that. Rather than testing the entire design page, I structured the test to take screenshots of each iteration of the component on the page so they appear as separate steps in Applitools and allow you to pass and fail individual views rather than full page of them which would be harder to reason.

https://applitools.com/blog/why-screenshot-image-comparison-tools-fail
https://applitools.com/blog/how-to-do-image-comparison-right

myspivey:
> Thanks for the response. Indeed, Chromatic only supports storybook. I didn't see that at first 😬
> 
> Can we see how the test results look like etc? I'm convinced we need a visual test tool, but still have to look at alternatives.

Absolutely! You can signup here, https://applitools.com/users/register. Once you are signed up, you can use the provided code to run the tests. Just export it to your terminal profile `export APPLITOOLS_API_KEY={MY_KEY}`. Any test runs you make locally will show up on your dashboard. The limits are generous enough that this should get you up and running. If you want to take it to the next step of getting the OSS license for builds, just let me know!

luisrudge:
Thanks for the PR. We hope we can find the time to investigate and adopt one of those tools, but now it's not the time 😬

PR #1605, opened 2019-02-20
fix(i18n): use 您 for consistency

blackwindforce:


luisrudge:
Can you explain what this change does?

blackwindforce:
This change uses formal "you" in Chinese, and keep using it consistently.

zachguo:
@luisrudge 您 is politer than 你, and is usually used in formal and written articulation.
@blackwindforce Can you please also do the same for the simplified Chinese file?

luisrudge:
Thanks for the PR 🎉

PR #1624, opened 2019-03-27
Add new telemetry information

luisrudge:
### Changes

Added new telemetry format

### Testing


src/__tests__/core/web_api/p2_api.test.js
```diff
@@ -37,7 +38,22 @@ describe('Auth0APIClient', () => {
   });
   describe('init', () => {
     describe('with overrides', () => {
+      it('uses telemetry set in the `auth0Client` query param if available', () => {
+        const telemetryIn = { fromTest: true, env: { envOverride: true } };
```
lbalmaceda: @luisrudge I think we should whitelist the params we allow in the telemetry. e.g. only allow to change the ones we will need on the backend. @joshcanhelp Any strong feelings on this or just leave it as it is?

src/core/web_api/p2_api.js
```diff
@@ -17,11 +18,24 @@ class Auth0APIClient {
     this.isUniversalLogin = window.location.host === domain;
     this._enableIdPInitiatedLogin = !!(opts._enableIdPInitiatedLogin || opts._enableImpersonation);
 
-    const default_telemetry = {
-      name: 'lock.js',
-      version: __VERSION__,
-      lib_version: auth0.version
-    };
+    const { auth0Client } = qs.parse(window.location.search.substr(1));
```
lbalmaceda: won't this include anything on the query, not just the auth0Client query param?

luisrudge: this is parsing the query string and getting only the `auth0Client` key

src/__tests__/core/web_api/p2_api.test.js
luisrudge: this wasn't in the RFC, so it's not here 😬 ideally, people should be able to use whatever telemetry they want, right?

src/core/web_api/p2_api.js
lbalmaceda: Ah, I just see the destructuring 👍

src/__tests__/core/web_api/p2_api.test.js
joshcanhelp: > ideally, people should be able to use whatever telemetry they want, right?

Ideally, no, but I'm not sure how far we need to go towards ensuring what is sent. I, personally, don't like the idea of opening up the telemetry to whatever they want to send, would rather it restrict to what's parsed (`name`, `version`, `env`) but processing on the data side will only look at those properties.

Technically, anyone **CAN** send anything they want as a header/query param outside of our SDKs. To be honest, I don't know how that will appear in the logs. In the actual reports, though, only specific properties will appear.

All of that said ... I would not use untracked properties in the tests, no reason to do that. Just set a `name` of `test-sdk` or similar and a version. If for some reason we whitelist properties, this would fail unnecessarily. Also, bad example if someone is grepping for examples.

luisrudge: @joshcanhelp I updated the test props.

joshcanhelp:
@luisrudge - So ... before we merge this in, I think we should resolve the "telemetry on the ULP" question. As in ... should we just nuke telemetry when Lock/Auth0.js is used on our server? I tend to lean towards "yes" here.

I'll post this internally for discussion 👍

PR #1627, opened 2019-04-04
Changes german translation of 'sign up' from 'Anmelden' to 'Registrieren;

PapaMufflon:
'Sign up' in German is 'Registrieren', not 'Anmelden'.


lbalmaceda:
@PapaMufflon That's for the title of the page, not the action button itself. Is this still correct? I'll check internally as well.

PR #1648, opened 2019-05-08
Fix telemetry

luisrudge:
### Changes

Fix telemetry with overrides and inside ULP

### DIFF
--- a/src/__tests__/core/web_api/p2_api.test.js
+++ b/src/__tests__/core/web_api/p2_api.test.js
@@ -1,5 +1,4 @@
 import { setURL } from 'testUtils';
-import { watch } from 'fs';
 
 jest.mock('auth0-js');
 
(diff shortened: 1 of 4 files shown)


src/__tests__/core/web_api/p2_api.test.js
```diff
@@ -86,7 +102,6 @@ describe('Auth0APIClient', () => {
               name: 'ExamplePlugin'
             }
           ],
-          _telemetryInfo: { newOption: true },
```
lbalmaceda: shouldn't you keep this to show that it's being ignored?

src/core/web_api/p2_api.js
```diff
@@ -82,6 +60,50 @@ class Auth0APIClient {
...
+        name: 'lock.js-ulp',
+        version: getVersion(),
+        env: {
+          ['auth0-js-ulp']: auth0.version.raw
+        }
+      };
+    }
+    if (!this.isUniversalLogin && telemetryOverride) {
+      telemetry = {
+        ...telemetryOverride,
```
lbalmaceda: what if this already has an `env` property? how would that behave? I would like it to me merged

luisrudge: see two lines down

src/__tests__/core/web_api/p2_api.test.js
luisrudge: I removed from here because I tested all the scenarios in the other test, so there's no point in validating the telemetry here again

PR #1653, opened 2019-05-22
Fix destroying lock instance

### LINKED ISSUE #907: Unable to destroy Lock instance

luisrudge:
fix #907, https://github.com/auth0/lock/issues/1090, https://github.com/auth0/lock/pull/1091

### Changes

This is a copy of https://github.com/auth0/lock/pull/1091, but I couldn't rebase because the original branch was removed.

This fixes a case where lock couldn't be destroyed `lock.destroy()`.

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -7,7 +7,9 @@ export function observe(key, id, f) {
   subscribe(`${key}-${id}`, (_, oldState, newState) => {
     const m = getEntity(newState, 'lock', id);
     const oldM = getEntity(oldState, 'lock', id);
-    if (m != oldM) f(m);
+    if (m && !m.equals(oldM)) {
+      f(m);
+    }
   });
 }
 


src/store/index.js
```diff
@@ -7,7 +7,9 @@ export function observe(key, id, f) {
   subscribe(`${key}-${id}`, (_, oldState, newState) => {
     const m = getEntity(newState, 'lock', id);
     const oldM = getEntity(oldState, 'lock', id);
-    if (m != oldM) f(m);
+    if (m && !m.equals(oldM)) {
```
damieng: Perhaps (m != null) would be safer?  What types does getEntity return?

luisrudge: they will always be different because `m` is an immutable object (which is a new reference, so will always be false if you compare with !=). that's why the fix is actually use `.equals`, which is an immutable method.

PR #1656, opened 2019-05-24
Add support for root level attributes

luisrudge:
### Changes

When using `additionalSignUpFields`, you can now use a new property called `storage` to determine where the field will be stored. `root` will save it in the root profile. anything else will save it under `user_metadata`. There is no default (which means it will save to `user_metadata`).

### References

https://auth0.com/docs/api/management/v2#!/Users/patch_users_by_id

### Testing


luisrudge:
@joshcanhelp made it opt-in per field

damieng:
@joshcanhelp it doesn't change behavior for existing fields - right now they go to user metadata.  The default for 'storage' is user metadata.  

The rationale for this design was to:

1. Keep backwards compatible
2. Allow for future expansion if other alternatives to user metadata and root profile attributes are added
3. Make it clear where data is stored

(A flag or opt-in to root profile attributes tied to that name, e.g. useRootProfileAttributes - fails objective 2 and 3)

luisrudge:
@joshcanhelp updated PR and description

PR #1657, opened 2019-05-29
Add validation to new root profile attributes

luisrudge:
### Changes

Add default validation for new root profile attributes

### Testing


src/field/index.js
```diff
@@ -4,15 +4,31 @@ import trim from 'trim';
...
-export function setField(m, field, value, validator = str => trim(str).length > 0, ...args) {
+const minMax = (value, min, max) => value.length >= min && value.length <= max;
+
+const getDefaultValidator = field => {
+  switch (field) {
+    case 'family_name':
+      return str => minMax(trim(str), 1, 150);
+    case 'given_name':
+      return str => minMax(trim(str), 1, 150);
+    case 'name':
```
damieng: Nitpick: You could write this as:

```javascript
switch (field) {
  case 'family_name':
  case 'given_name':
    return str => minMax(trim(str, 1, 150);
```

etc.

src/field/username.js
```diff
@@ -55,5 +55,5 @@ export function setUsername(m, str, usernameStyle = 'username', validateUsername
 }
 
 export function usernameLooksLikeEmail(str) {
-  return str.indexOf('@') > -1;
+  return str.indexOf('@') > -1 && str.indexOf('.') > -1;
```
damieng: Wonder if we should check the '.' comes after the @ here?

```javascript
return str.indexOf('.') > str.indexOf('@') && str.indexOf('@') > -1;
```

support/index.html
```diff
@@ -126,7 +126,7 @@ <h1 class="navbar-brand">
         usernameStyle: 'email',
         defaultDatabaseConnection: 'acme',
         prefill: {
-          email: 'johnfoo@gmail.com'
+          email: 'luis@luisrudge.net'
```
damieng: Probably going to get a lot more spam from web crawlers...  But then johnfoo might appreciate it.

src/field/username.js
luisrudge: this would return false for `test.email@t.com`

support/index.html
luisrudge: fixed

src/field/index.js
luisrudge: fixed

src/field/username.js
```diff
@@ -3,8 +3,8 @@ import { validateEmail } from './email';
 import { databaseConnection } from '../connection/database';
 import trim from 'trim';
 
-const DEFAULT_CONNECTION_VALIDATION = { username: { min: 1, max: 15 } };
-const regExp = /^[a-zA-Z0-9_+\-.]+$/;
+const DEFAULT_CONNECTION_VALIDATION = { username: { min: 1, max: 128 } };
```
panga: The default didn't change, still 15 chars.

src/field/username.js
panga: Can you please use a more robust email validation library?
Ideally, the same library used in the backend should be used in Lock: 
https://github.com/validatorjs/validator.js#client-side-usage

src/__tests__/field/username.test.js
```diff
@@ -134,11 +136,11 @@ describe('field/username', () => {
             expectToFailWith('aaaaaa');
           });
           it('validates invalid chars', () => {
-            const invalidChars = `{}[],;?/\\!@#$%¨&*()¹²³\`~^´ªº§£¢¬<>|"' `.split('');
+            const invalidChars = `{}[],;?/\\%¨&*()¹²³ªº§£¢¬<>|" `.split('');
             invalidChars.forEach(i => expectToFailWith(`aa${i}`));
           });
           it('accepts letters, numbers, `_`, `-`, `+` and `.`', () => {
-            const validChars = `_-+.`.split('');
+            const validChars = `_+-.!#$'^\`~@`.split('');
```
panga: Suggestion: Update test title with new chars

src/field/username.js
luisrudge: this validation is only for cases that the browser doesn't support html5 validation, since we use that by default. we're not adding +20kb just to do email validation 😬

PR #1681, opened 2019-06-28
use --trace when uploading to the cdn

luisrudge:
### Description

This will make the logs in our build server to display more information about the cdn upload

(the package was also auto formatted)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,15 @@
   "description": "Auth0 Lock",
   "author": "Auth0 <support@auth0.com> (http://auth0.com)",
   "license": "MIT",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -45,7 +50,7 @@
     "babel-preset-stage-0": "^6.3.13",
     "babelify": "^7.2.0",
     "bump-version": "^0.5.0",
-    "component-cdn-uploader": "auth0/component-cdn-uploader#v1.3.0",
+    "component-cdn-uploader": "auth0/component-cdn-uploader#v2.2.1",
```
lbalmaceda: are there breaking changes after this change in the major?

PR #1701, opened 2019-07-18
Fix call to undefined function cb

### LINKED ISSUE #1700: ReferenceError: cb is not defined
* Lock version: 11.17.0 (possibly since 10.0.0)

There is an undefined function `cb()` called.

Was introduced with: https://github.com/auth0/lock/commit/7ea5c2271a63f6b09b19443ceddb4a17bc48adbb#diff-d9c9f2bcdfe158127d37c0b3e33e8d0eR168

image

KevinBusse:
Fixes #1700

### Changes

Remove call to undefined function `cb`.
No visual changes  - except no error in console.

### DIFF
--- a/src/ui/box/multisize_slide.js
+++ b/src/ui/box/multisize_slide.js
@@ -172,7 +172,6 @@ class Child extends React.Component {
   componentWillUnmount() {
     if (this.t) {
       clearInterval(this.t);
-      cb();
     }
   }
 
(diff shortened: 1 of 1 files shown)


KevinBusse:
After intensely staring on the commit introducing the issue I assume that simply removing `cb()` is the wrong move.

`this.cb()` should have been called instead. In that case please adopt the code to fit its needs.

Issue #1703, opened 2019-07-19
Unable to install via npm due to invalid component-cdn-uploader


Enngage:
Hi,

Today I started getting following error when installing `"auth0-lock": "11.17.0"`:

```
npm WARN deprecated core-js@1.2.7: core-js@<2.6.8 is no longer maintained. Please, upgrade to core-js@3 or at least to actual version of core-js@2.
npm ERR! path git
npm ERR! code ENOENT
npm ERR! errno ENOENT
npm ERR! syscall spawn git
npm ERR! enoent Error while executing:
npm ERR! enoent undefined ls-remote -h -t ssh://git@github.com/auth0/component-cdn-uploader.git
npm ERR! enoent
npm ERR! enoent
npm ERR! enoent spawn git ENOENT
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent
```

I tried installing `auth0-lock` dependency on https://stackblitz.com, but got same result: 

![asefase](https://user-images.githubusercontent.com/7137351/61567209-11571900-aa7f-11e9-9e19-b13342bd7cd8.png)

Can someone shed some light on this? 

Thanks!

luisrudge:
Sorry! There was two errors here: first, this dependency should be a development dependency. The other one is that we should load if via npm and not git. I opened a PR to fix this and we'll have this fixed in a patch fix on Monday!

luisrudge:
Fixed in 11.17.1

PR #1704, opened 2019-07-19
Fix CDN dependency

luisrudge:
fix https://github.com/auth0/lock/issues/1703

### Changes

For some reason, we were using the github version of the lib and this fails in some cases (when the machine doesn't have git installed etc). This PR changes the dependency to load from npm.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,9 +2,10 @@
 # yarn lockfile v1
 
 
-"@auth0/component-cdn-uploader@auth0/component-cdn-uploader#v2.2.2":
+"@auth0/component-cdn-uploader@^2.2.2":
   version "2.2.2"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -662,9 +663,9 @@ autoprefixer@^6.3.1:
     postcss-value-parser "^3.2.3"
 
 aws-sdk@^2.346.0:
-  version "2.484.0"
-  resolved "https://registry.yarnpkg.com/aws-sdk/-/aws-sdk-2.484.0.tgz#ec1039f0fae688eb120729ca35f0c686043199df"
-  integrity sha512-zGWbN9okRRSXvJDYXLdPGuSri+w6X7Fp5MxztlkXWsS6GCm7Lya8Iiti1cYa3/S4PgprmScHSCk+sFcwGRLBfw==
+  version "2.496.0"
```
damieng: Is this meant to be in this PR as well?

PR #1712, opened 2019-08-03
Warn when `auth.redirectUri` is sent as an option

luisrudge:
### Changes

Make sure devs copying/pasting code from auth0.js get a warning because both options have very similar names

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -251,7 +251,6 @@ function extractAuthOptions(options) {
     params,
     autoParseHash,
     redirect,
-    redirectUri, //invalid prop. it's only here to show a warning
     redirectUrl,
     responseMode,
     responseType,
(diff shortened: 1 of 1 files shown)


charlesjohnson:
Instead of exiting 0 with a “WARN: Your script will exit 0 and fail silently,” should it just exit 1 instead?

luisrudge:
@charlesjohnson we can't fail silently now without a breaking change. The warn is the best we can do without breaking existing users

src/core/index.js
```diff
@@ -251,14 +251,19 @@ function extractAuthOptions(options) {
...
     redirectUrl,
     responseMode,
     responseType,
     sso,
     state,
     nonce
-  } =
-    options.auth || {};
+  } = options.auth || {};
+  if (redirectUri) {
```
lbalmaceda: I'd keep the previous approach that didn't require you to expose a new variable to this function's scope. (rollback that last commit). Maybe keep the comment inside this IF clause @luisrudge

luisrudge: done

Issue #1719, opened 2019-08-22
`password_expired` error during login is not being mapped to `password_change_required`


xiananfan:
**Thank you for using Lock!**

In order to efficiently and accurately address your issue or feature request, please read through the template below and answer all relevant questions. Your additional work here is greatly appreciated and will help us respond as quickly as possible. Please delete any sections or questions below that do not pertain to this request.

For general support or usage questions, please use the [Auth0 Community](https://community.auth0.com/) or [Auth0 Support](https://support.auth0.com.).

### Description

Description of the bug or feature request and why it's a problem. Consider including:

- The use case or overall problem you're trying to solve
- Information about when the problem started

Hi Auth0 team,

We are using AD connector on a windows server to talk to our Active Directory. And we have a 90 days password expired policy. When a user's password is expired, the lock widget displays an unclear message: "We're sorry, something went wrong when attempting to log in."

After looking into the code [here](https://github.com/auth0/lock/blob/fb4948c38808e8ef900deb393840f76090ff4864/src/i18n/en.js#L13), it seems like the lock is using the message for "lock.fallback", it probably should be displaying the message for "password_change_required".

When looking at this [method](https://github.com/auth0/lock/blob/master/src/core/index.js#L497), it does not map any message for "password expired", as a result, it will display "lock.fallback" message. Please check for password expired error as well, and display the correct message.

Thanks,
Xianan

### Prerequisites

* [x] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
* [x] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)
* [x] Did you check the documentation in the [README](https://github.com/auth0/lock) or on [our docs site](https://auth0.com/docs/libraries/lock/v11)?
* [x] Did you check the [Auth0 Community](https://community.auth0.com/)?
* [x] Are you reporting this to the correct repository?
* [x] Are there any related or duplicate [Issues](https://github.com/auth0/lock/issues) or [PRs](https://github.com/auth0/lock/pulls) for this issue?

### Environment

Please provide the following:

* Lock version: 11.16.0
* Browser version: Chrome 76.0.3809.100 
* OS: MacOS 10.14.6
* Additional extensions/libraries that might be affecting your instance:

### Reproduction

Detail the steps taken to reproduce this error and note if this issue can be reproduced consistently or if it is intermittent.

[...]

PR #1730, opened 2019-09-24
Map password_expired to password_change_required

luisrudge:
### Changes

This PRs fixes an error mapping (from `password_expired` to `password_change_required`) in order to provide a better error message for end users.

### References

Fix https://github.com/auth0/lock/issues/1719

### Testing


joshcanhelp:
@luisrudge - Looks like this is a WIP so I removed our team for now. Add us back when you're ready!

PR #1735, opened 2019-10-03
Relaxing the email validation

luisrudge:
### Changes

To better handle unicode values in the domain part, we need to relax the email validation. After some internal discussions, I decided that it's easier to let the server decide if the email is valid or not. This PR only validates that the email:
- is a string
- has a `@`
- has a `.`
- doesn't have a ` ` (white space)

I think this is all that is right amount of balance between the user experience and actual email validation

### References

ESD-2239

### Testing

### DIFF
--- a/src/__tests__/field/email.test.js
+++ b/src/__tests__/field/email.test.js
@@ -6,12 +6,14 @@ describe('field/email', () => {
       expect(isEmail('test@test.com')).toBe(true);
       expect(isEmail('test@test.com.br')).toBe(true);
       expect(isEmail('test@移动.移动')).toBe(true);
+      expect(isEmail('test@test.com ')).toBe(true);
+      expect(isEmail(' test@test.com')).toBe(true);
       expect(isEmail(1)).toBe(false);
       expect(isEmail('test@testcom')).toBe(false);
       expect(isEmail('test.test.com')).toBe(false);
     });
-    it('returns false when there is a white space', () => {
-      expect(isEmail('test@test.com ')).toBe(false);
+    it('returns false when there is a white space in the middle of the string', () => {
+      expect(isEmail('test@test. com')).toBe(false);
     });
   });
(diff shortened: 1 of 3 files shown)


src/field/email.js
```diff
@@ -1,18 +1,18 @@
...
 import * as i18n from '../i18n';
 
-const regExp = /^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
-
 export function validateEmail(str) {
   return isEmail(str);
 }
 
 export function isEmail(str) {
-  const result = regExp.exec(trim(str.toLowerCase()));
```
lbalmaceda: I think you should keep the "trim" (removes spaces on edges of the string) to conserve the old behavior.

src/__tests__/field/email.test.js
```diff
@@ -0,0 +1,27 @@
...
+    it('validates correctly', () => {
+      expect(isEmail('test@test.com')).toBe(true);
+      expect(isEmail('test@test.com.br')).toBe(true);
+      expect(isEmail('test@移动.移动')).toBe(true);
+      expect(isEmail(1)).toBe(false);
+      expect(isEmail('test@testcom')).toBe(false);
+      expect(isEmail('test.test.com')).toBe(false);
+    });
+    it('returns false when there is a white space', () => {
+      expect(isEmail('test@test.com ')).toBe(false);
```
lbalmaceda: This is a valid scenario if trim is enabled. However, a space in between should be invalid.

src/field/email.js
luisrudge: fixed

PR #1888, opened 2020-06-23
[CAUTH-500] Reload recaptcha after wrong username or password

jfromaniello:
### Changes

I've refactored our ReCaptcha component in a more react-ish class. I've tested this in deep and now it seems to work as we expect.

This should be released as a **patch** because we are not changing any existing API nor adding a new ones.

I am also attaching in this PR a couple of images showing the wrong behavior.

## Wrong behavior

Screen-Recording-2020-06-24-16-28-41

**As you can see after getting the error the recaptcha can't be re-checked, and the form can't be submitted.**

## Fixed behavior 
Screen-Recording-2020-06-24-16-04-55

### References

https://auth0team.atlassian.net/browse/CAUTH-500

### Testing

* [x] This change adds integration test coverage

### Checklist

### DIFF
--- a/src/__tests__/field/captcha/recaptchav2.test.jsx
+++ b/src/__tests__/field/captcha/recaptchav2.test.jsx
@@ -2,13 +2,13 @@ import React from 'react';
 import { shallow } from 'enzyme';
 import I from 'immutable';
 
-import RecaptchaV2, { render } from '../../../field/captcha/recaptchav2';
+import { ReCAPTCHA } from '../../../field/captcha/recaptchav2';
 
-const createLockMock = ({ provider = 'none', siteKey = '' } = {}) =>
+const createLockMock = ({ provider = 'none', sitekey = '' } = {}) =>
   I.fromJS({
     id: '__lock-id__',
     core: {
-      captcha: { provider, siteKey },
+      captcha: { provider, sitekey },
       transient: {
         ui: {
           language: 'en-US'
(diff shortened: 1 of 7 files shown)


src/__tests__/field/captcha/recaptchav2.test.jsx
```diff
@@ -32,10 +33,10 @@ describe('RecaptchaV2', () => {
       document.getElementById('renderTest').remove();
     });
     it('injects the script', () => {
-      const mockLock = createLockMock({ provider: 'recaptchav2', siteKey: 'mySiteKey' });
-      render(mockLock, document.getElementById('renderTest'), {});
-      expect(document.body.innerHTML).toBe(
-        '<div id="renderTest"></div><script src="https://www.google.com/recaptcha/api.js?hl=en-US"></script>'
+      ReCAPTCHA.loadScript({ hl: 'en-US' }, document.body);
```
lbalmaceda, line 36: @stevehobbsdev language is the same as the line above, but could this support the language lock is configured to?

src/field/captcha/captcha_pane.jsx
```diff
@@ -7,22 +7,38 @@ import * as l from '../../core/index';
...
+      function reset() {
+        handleChange();
+      }
+
+      return (
+        <ReCAPTCHA
+          sitekey={captcha.get('siteKey')}
+          onChange={handleChange}
+          onExpired={reset}
+          hl={l.ui.language(lock)}
```
lbalmaceda, line 34: Here we're taking it from Lock options. I wonder if the line with the fixed language I pointed in my other comment has any effect on this.

jfromaniello: The other comment is in a line inside a unit testing where I am testing that the recaptcha component links to the right url in google. Then in this line we are instantiating the recaptcha component with the language that was used in Lock. I've tested this manually and it works as expected, if lock is configured in spanish then recaptcha is in spanish, if lock is in english (or the lang is not set) then recaptcha is in English. 

If you don't specify the language to google recaptcha it will default to the language of your browser, which might lead to a broken ui. So I tested that we always send the language to google

src/__tests__/field/captcha/recaptchav2.test.jsx
jfromaniello: This is a unit test only of the recaptcha component, when this component is instantiated whitin the "captchapane" the language sent is the lang used to configure lock

src/field/captcha/captcha_pane.jsx
lbalmaceda: Thanks. I completely missed the other was part of a unit test. 🤕

PR #2156, opened 2022-07-21
Fixed add captcha to passwordless

sdwvit:
Temporary pr for POC, will close by the end of this week


src/engine/matchesEnterpriseConnection.js
```diff
@@ -0,0 +1,11 @@
+import {isEnterpriseDomain} from "../connection/enterprise";
```
sdwvit, line 1: file probably needs better name

PR #2222, opened 2022-12-14
IAMRISK-1790 Support captcha for Passwordless

robinbijlani:
### Changes

This adds support for captcha in Passwordless flows (Classic Universal Login).

For context, to allow this, the Authentication API has [some changes](https://github.com/auth0/auth0-server/pull/10632):

- `POST /passwordless/challenge` can be hit to know if a captcha should be shown for Passwordless flows.
- `POST /passwordless/start` will now verify the `captcha` field if one is required.

Here are a few demo recordings of the behavior:

passwordless-email-auth0

passwordless-email-recap

passwordless-sms-auth0
☝️ This recording shows the one clunky UX issue I found. When using `restart()` to go back to the first pane, I can't figure out how to clear out the `captcha` field from the model. (I tried several implementations of `('../../field/index').clearFields(m, ['captcha']);` combined with `swap()` but I am think I'm somewhat foggy on how the model binding works here.)

passwordless-sms-recap

### References

https://auth0team.atlassian.net/browse/IAMRISK-1790

### Testing

### Checklist


src/connection/passwordless/actions.js
```diff
@@ -28,15 +28,25 @@ function getErrorMessage(m, error) {
     key = 'bad.phone_number';
   }
 
+  if (error.code === 'invalid_captcha') {
+    const captchaConfig = l.passwordlessCaptcha(m);
+    key = captchaConfig.get('provider') === 'recaptcha_v2' ? 'invalid_recaptcha' : 'invalid_captcha';
```
ktraff-auth0, line 33: Do we need to think about recaptcha enterprise here too?

src/connection/passwordless/actions.js
```diff
@@ -122,9 +142,10 @@ function sendSMSSuccess(id) {
 
 function sendSMSError(id, error) {
   const m = read(getEntity, 'lock', id);
-  const errorMessage = getErrorMessage(m, error);
+  const errorMessage = getErrorMessage(m, id, error);
   l.emitAuthorizationErrorEvent(m, error);
-  return swap(updateEntity, 'lock', id, l.setSubmitting, false, errorMessage);
+  swap(updateEntity, 'lock', id, l.setSubmitting, false, errorMessage);
```
ktraff-auth0, line 147: I didn't look at this in detail but this previous returned the value from `swap()`, does the fact that this function returns void now present any issues?

src/core/index.js
```diff
@@ -430,6 +436,13 @@ export function captcha(m) {
   return get(m, 'captcha');
 }
 
+export function passwordlessCaptcha(m) {
+  if (typeof m !== 'object') {
```
ktraff-auth0, line 440: Is this check necessary?  Should get handle non-objects (just thinking of lodash)?

src/i18n/af.js
```diff
@@ -35,7 +35,9 @@ export default {
     passwordless: {
       'bad.email': 'Die e-pos is ongeldig',
       'bad.phone_number': 'Die telefoonnommer is ongeldig',
-      'lock.fallback': 'Jammer, iets het verkeerd gegaan'
+      'lock.fallback': 'Jammer, iets het verkeerd gegaan',
+      invalid_captcha: "Los die uitdagingsvraag om te verifieer dat u nie 'n robot is nie.",
+      invalid_recaptcha: "Kies die merkblokkie om te verifieer dat u nie 'n robot is nie."
```
ktraff-auth0, line 40: Wow you know many languages!

yarn.lock
```diff
@@ -850,9 +850,9 @@ atob@^2.1.2:
   resolved "https://registry.yarnpkg.com/atob/-/atob-2.1.2.tgz#6d9517eb9e030d2436666651e86bd9f6f13533c9"
   integrity sha512-Wm6ukoaOGJi/73p/cl2GvLjTI5JM1k/O14isD73YML8StrH/7/lRFgmg8nICZgD3bZZvjwCGxtMOD3wWNAu8cg==
 
-auth0-js@^9.19.2:
+auth0-js@^9.20.0:
   version "9.20.0"
-  resolved "https://registry.yarnpkg.com/auth0-js/-/auth0-js-9.20.0.tgz#80d140a6741f677f5295f2b38e750e23961726fe"
+  resolved "https://a0us.jfrog.io/a0us/api/npm/npm/auth0-js/-/auth0-js-9.20.0.tgz#80d140a6741f677f5295f2b38e750e23961726fe"
```
stevehobbsdev, line 855: We can't use dependencies from Artifactory as no-one outside of the organization will be able to install it, it's also causing the Circle build to fail.

src/connection/passwordless/actions.js
stevehobbsdev: We should be ok in this case, the only caller to `sendSMSError` discards the return value

https://github.com/auth0/lock/blob/872f7f767ce02cabf753f3e0672fe3fcc3602fa8/src/connection/passwordless/actions.js#L107

stevehobbsdev:
> I can't figure out how to clear out the captcha field from the model. (I tried several implementations of ('../../field/index').clearFields(m, ['captcha']); combined with swap() but I am think I'm somewhat foggy on how the model binding works here.)

@robinbijlani there is [a `reset` helper](https://github.com/auth0/lock/blob/872f7f767ce02cabf753f3e0672fe3fcc3602fa8/src/core/index.js#L420) on the captcha field that should allow you to reset it, did you try this one?

This is used on Captcha for non-passwordless:

```js
import * as captchaField from '../field/captcha';

// ....

export function setCaptcha(m, value, wasInvalid) {
  m = captchaField.reset(m, wasInvalid);
  return set(m, 'captcha', Immutable.fromJS(value));
}
```

src/connection/passwordless/actions.js
robinbijlani: Thank you for the callout. Since `recaptcha_v2` and `recaptcha_enterprise` both offer the same UX, I've made the error messages match everywhere we are showing captcha. 👍

src/core/index.js
robinbijlani: I copied it from the `captcha()` function above, but it looks like this isn't needed for the current tests. 🤔 For now I removed it from both, but maybe @stevehobbsdev can double check this.

yarn.lock
robinbijlani: Of course. 🤦‍♀️

robinbijlani:
@stevehobbsdev Thank you for the guidance! I realized my mistake was that while we have a `passwordlessCaptcha` prop on the model (to store info about the captcha challenge we display), we have no separate _field_ for the passwordless captcha, and are reusing the existing captcha one. Now things seem to be behaving like I would expect:

passwordless-email-auth0-v2
passwordless-email-recap-v2

robinbijlani:
Also, the Browserstack failure is expected, right? 🤔 
```
No browserstack username!
No browserstack password!
```

robinbijlani:
@stevehobbsdev Also, I know you're OOF until Monday, but when you get back, if this looks good I'd love for a new release of Lock to be cut. (I want to bump the default version used from auth0-server.)

stevehobbsdev:
> Also, the Browserstack failure is expected, right?

@robinbijlani Yes, as these changes come from a fork, the Browserstack run does not get the secrets required to execute. I will check the E2E tests locally.

stevehobbsdev (approved):

PR #2371, opened 2023-06-26
Update index.styl

acolbourn:
Update box shadow

### Testing

### Checklist

### DIFF
--- a/css/index.styl
+++ b/css/index.styl
@@ -1024,7 +1024,7 @@ loadingSize = 30px
         transition transform .6s, opacity .6s
         -webkit-transition-delay .5s
         transition-delay .5s
-        box-shadow 0 0 40px 4px #111118
+        box-shadow 0px 10px 15px -3px rgba(0,0,0,0.35)
 
       // mobile view
       @media screen and (max-width: 480px)


acolbourn:
The box shadow just looked way different than the preview and looked weird, this isn't critical if this is too complicated to update.

stevehobbsdev:
@acolbourn Can you provide a screenshot showing the differences?

stevehobbsdev:
Also this could be overridden in your own app, or in the login page template if you're using Auth0 Universal Login - does either of those options work for you?

acolbourn:
Here is what the preview menu shows the login will look like: 
Preview login

And here is what the actual login ends up looking like:
Actual Login

In the customization tool I can turn off the box shadow completely but can't edit the box shadow directly which is why I made this PR (I actually thought I was just editing my own local copy when I made this PR).  Anyway, this really isn't too big a deal, it just wasn't clear why the actual login looks completely different from the preview or how I can customize my own version.  I suspect I just did something wrong so maybe we just close this PR.

stevehobbsdev:
Ah thanks for sharing the screenshots. Those are actually [two different login experiences](https://auth0.com/docs/authenticate/login/auth0-universal-login#choose-login-experience) we have in Auth0 ("New" and "Classic"), which I appreciate can be confusing.

The first one (with the Welcome heading) is the New experience and this is what the preview shows. This is a completely hosted experience and you're able to [customize it to an extent with in the Auth0 Dashboard](https://auth0.com/docs/authenticate/login/auth0-universal-login#new-universal-login-experience).

The second one is the "Classic" experience that is more or less legacy at this point, and this library powers that UI widget that your users interact with. You can use it either hosted (the classic experience) or you can self-host embedded it in your own projects.

You would normally want to use the New hosted experience unless you really need an embedded experience, but there you're able to customize the UI anyway.

Anyway, that's why the two login boxes look different. I'll close this PR for now then but happy to continue the discussion, and do let me know if you have any further questions 👍🏻

acolbourn:
Ok great thanks, we'll see if we can swap over to the new one.

ConnieZi:
@stevehobbsdev Based on my experience so far, since we have to use lock in our signin widget, the default look is the legacy one, and it does't support much customization in its UI. 

I modified the signin widget by going through Universal Login-> Login->Default Templates(selected Lock), and then the default (pre)view was the legacy one. In the docs it also recommends us not to overwrite the CSS and only limits the customization to functional parts instead of UI parts.

PR #2374, opened 2023-06-29
Email and Password input svgs - text alternative is missing

piwysocki:
### Changes

Title tags added to email and password input SVGs.

It's an accessibility issue: The text alternative is missing for that icons, so screen reader users won't be able to understand the  purpose of the icons.

### Testing

### Checklist


frederikprijck (changes requested):
Thanks for the PR.

Can you elaborate on the a11y benefits for **non-focusable** SVGs here that are marked with **aria-hidden=true**? These should be skipped by screenreaders anyway, so I am not sure I understand the benefit of the change being proposed. 

Regardless, It's always helpful to provide context, both for anyone reviewing but as well for anyone looking at this PR in the future.

Thanks!

piwysocki:
Image must have a text alternative, so blind people can get access to information carried by the image. Text alt should describe information presented by image. Screen readers can use it to pass information to the user.

frederikprijck:
> Screen readers can use it to pass information to the user.

The svgs are marked with `aria-hidden=true` to hide it from screen readers or any accessibility API. So I am trying to understand the intention here.

piwysocki:
Visually impaired person can hover mouse over an image and see a text with an input description.

frederikprijck:
For visually impaired and blind people, or anyone using screen-readers, we provide aria-labels on the input elements, and all kind of other attributes to help them. The icon is purely decorative, which is why it's marked with `aria-hidden` set to true.

Given this PR isn't mergable in it's current state (we use i18n, not hard coded abels) and I don't see the benefit, I'm going to close this PR. If you strongly believe we should add this I would recommend providing some additional information that indicates the need for this change, including some a11y reporting that indicates the a11y issue here.

pbuckleyfen:
To adhere to the WCAG Rule: 1.1.1.b Alternative Text (Informative Images)- Images that convey content MUST have programmatically-discernible alternative text.

As there is no text when a email and password is populated, Alternative text to an aria-label should be provided. 

Multiple audits of this screen by accessibility providers have flagged this as a critical accessibility issue

frederikprijck:
Thanks @pbuckleyfen , that same rule also says:

> If non-text content is pure decoration, is used only for visual formatting, or is not presented to users, then it is implemented in a way that it can be ignored by assistive technology.

So I'm not sure i fully understand. What tools are u using that flags it as such? The svg is marked as aria hidden, so im not sure it should be flagged.

I will reopen this, but please provide a bit more information on the tools you used and the report it gave you so we can action this accordingly if needed.

frederikprijck:
I looked into this again, and ran the A11y checks using [Lighthouse](https://pagespeed.web.dev/) and [Axe](https://www.deque.com/axe/).

**Lighthouse:** 
image

You can try this yourself by pointing https://pagespeed.web.dev/ to your login page.

**Axe**:

image

You can try this yourself by installing the Axe chrome extension from https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd.

There is definitely room to improve the a11y, but adding alternative text to an element that is marked with aria-hidden isn't going to help from what I can see.

Both input fields have an `aria-label` to describe the input field. The icons are decorative only and are marked to be hidden from screen-readers as they should read the `aria-label` on the inputs to assist the users.

Closing this again. If you believe there is an issue, please open an **issue** with the necessary details for us to be able to reproduce and discuss it.

PR #2764, opened 2026-04-01
chore: remove package-lock.json and update CI to use npm install

ankita10119:
### 🛠️ Changes

1. Removes `package-lock.json `from version control and adds it to `.gitignore`. 
2. Updates CI workflows to use `npm install` instead of `npm ci` and removes `cache: 'npm'` from `actions/setup-node` which requires a lockfile to function.
3. Pins `cheerio` to `1.0.0-rc.12` via npm `overrides`.

#### Cheerio Override
**Why:** `enzyme@3.11.0` (test-only dependency) was built against cheerio's rc-series internals. When `package-lock.json` was removed, npm resolved `cheerio` to `1.2.0` (stable), which restructured the package entirely, removing internal paths enzyme depends on and adding `undici` as a new transitive dependency incompatible with the Jest/jsdom test environment.

  **Without it:** Jest test suites fail with a cascade of errors , `Cannot find module 'cheerio/lib/utils'`, `ReferenceError: TextEncoder is not defined`, `ReferenceError: ReadableStream is not defined`, and `SyntaxError: Unexpected token 'export'`.

  **Justification:** The enzyme maintainer committed this exact fix ([`cafdb2b`](https://github.com/enzymejs/enzyme/commit/cafdb2b86ed8865527cdd6ae31c42593d3728ceb)) in December 2020 but never cut a release. `enzyme@3.11.0` remains the latest published version, carrying the unfixed `^1.0.0-rc.3` range. This override is the [community workaround](https://github.com/enzymejs/enzyme/commit/cafdb2b86ed8865527cdd6ae31c42593d3728ceb) for a known upstream issue that will never be released.

  **Effect on consumers:** `overrides` is a local package manager instruction and not published to npm, not bundled in build output, ignored entirely by consumers' own installs.


cschetan77:
Closing this PR in favor of #2789

PR #2783, opened 2026-04-16
fix(deps): remove `trim` dependency

gameroman:
Removes `trim` dependency to use native `String.prototype.trim` instead

PR #2808, opened 2026-06-03
Release v14.3.1

cschetan77:
**Removed**
- fix(deps): remove `trim` dependency [\#2783](https://github.com/auth0/lock/pull/2783) ([gameroman](https://github.com/gameroman))

Removes `trim` dependency in favor of native `String.prototype.trim`


cschetan77:
Closing this out. This change will be released as part of v15 major version.

